PR #4, opened 2018-01-20
Reset the initial pointer position to the scrolling location when exe…

blankmac:
…cuting two finger scroll.


alexandred:
Thanks!

PR #5, opened 2018-01-26
Remove "double" right click when long press routine executes.

blankmac:
Holdover from the previous version when a different method was used to execute a right click.  It was causing a right click to be executed twice in a row in the rewrite.  If the perception of a 'left' click prior to the right click persists let me know.  This needs to be removed regardless.


alexandred:
Thanks!

PR #6, opened 2018-07-12
OS-X-Voodoo-PS2-Controller Integration - Touchpad Control

the-darkvoid:
This pull request adds integration for `OS-X-Voodoo-PS2-Controller` in order to enable/disable the touchpad through the `VoodooI2CPrecisionTouchpadHIDEventDriver` driver.

The enabling/disabling is implemented in `VoodooI2CMultitouchHIDEventDriver` and thus can be used on any derived classes through specifying the `RM,deliverNotifications` property.

Additionally the touchpad is temporarily disabled (currently 0.5 seconds, configurable through `QuietTimeAfterTyping` to prevent accidental palm interactions while typing.
This does not apply when using modifier keys for selection (i.e. ctrl + click, windows + click).

The described functionality requires the following pull request for `OS-X-Voodoo-PS2-Controller`:
https://github.com/RehabMan/OS-X-Voodoo-PS2-Controller/pull/158

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -100,6 +100,10 @@
 			<string>VoodooI2CPrecisionTouchpadHIDEventDriver</string>
 			<key>IOProviderClass</key>
 			<string>IOHIDInterface</string>
+			<key>RM,deliverNotifications</key>
+			<true/>
+			<key>QuietTimeAfterTyping</key>
+			<integer>500000000</integer>
 		</dict>
 		<key>VoodooI2CHIDDevice Multitouch HID Event Driver</key>
 		<dict>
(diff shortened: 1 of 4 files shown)


alexandred:
Brilliant, happy to merge this as it is. One suggestion for userfriendliness is that I would change `QuietTimeAfterTyping` to milliseconds in the configuration and convert it to nanoseconds in the code. That can be added in as a PR later if you prefer.

Is `QuietTimeAfterTyping` the only configuration value you intend to support? (so I can add it to the documentation).

I will make sure that this change makes it to the next minor version of VoodooI2C which I will probably release once your other change gets merged into VoodooPS2.

Thanks!

the-darkvoid:
You can ignore the earlier commit, it was an approach without having to modify OS-X-Voodoo-PS2-Controller, but having the change being driven by OS-X-Voodoo-PS2-Controller changes resulted in much cleaner code.

I'll look into changing the timing to milliseconds as suggested.

PR #8, opened 2018-07-18
Ignore built-in trackpad when mouse or wireless trackpad is present

the-darkvoid:
VoodooI2CHID can now listen for USB or Bluetooth HID devices connecting to the system.

If the device is found to be a pointer device (mouse, trackpad etc), this will disable the built-in trackpad automatically.

Each VoodooI2CHID derived class (for example VoodooI2CPrecisionTouchpadHIDEventDriver), can use this with two new configuration values:
* ProcessUSBMouseStopsTrackpad
* ProcessBluetoothMouseStopsTrackpad

Besides the kext level configuration, the user preference from macOS is taken into account:
`System -> Preferences -> Accessibility -> Mouse & Trackpad -> Ignore built-in trackpad when mouse or wireless trackpad is present`

Though with Bluetooth there can be a slight delay before the disconnection is detected and the touchpad re-enabled, hence its functionality is configurable.

Touchpad enabling and disabling via the PrtScr button works at any time (if the proper VoodooPS2Keyboard driver is installed).

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -104,6 +104,10 @@
 			<true/>
 			<key>QuietTimeAfterTyping</key>
 			<integer>500</integer>
+			<key>ProcessUSBMouseStopsTrackpad</key>
+			<true/>
+			<key>ProcessBluetoothMouseStopsTrackpad</key>
+			<true/>
 		</dict>
 		<key>VoodooI2CHIDDevice Multitouch HID Event Driver</key>
 		<dict>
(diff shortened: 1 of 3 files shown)


alexandred:
Can you confirm whether or not the necessary PRs have been accepted in the VoodooPS2 project? Does there now exist a release of VoodooPS2 with which this PR is compatible?

alexandred:
Scratch that, I misunderstood how this was implemented, I'm happy to accept it as is, thanks!

PR #10, opened 2018-09-23
Default USB/BT mouse search to false

ben9923:
This is not necessary with MT2 emulation, as the OS will ignore input itself when this feature is set via SysPrefs.
Users with just single touch capability, or an old OS can enable this feature via info.plist.
(Should probably be documented in the VoodooI2C documentation)


alexandred:
Thanks!

PR #16, opened 2019-04-13
Fix warnings across the project

ben9923:
Applied many fixes to warnings emitted by cpplint and Xcode.
Updated the project to use Xcode 10.2 recommended settings.
Fixed some small memory leaks found by Xcode's static analyzer.


alexandred:
Thanks!

PR #17, opened 2019-06-09
Fix various memory and resource leaks to properly unload the kext

kprinssu:
This *does not* fixe https://github.com/alexandred/VoodooI2C/issues/187, but will help address potenial causes of it.

The kernel extension still fails to properly unload and I believe the cause is now in the core VoodooI2C project.


ben9923:
Merged!

PR #20, opened 2019-10-26
Correct IOCommandGate::commandSleep call

al3xtjames:
Same as https://github.com/alexandred/VoodooI2C/pull/224.


ben9923:
Thank you!

PR #26, opened 2020-04-01
Optimise Polling Mode by adjusting time interval between I2C readings.

lvs1974:
Set time stamp when touch pad is used (in VoodooI2CMultitouchHIDEventDriver::handleDigitizerReport).
In VoodooI2CHIDDevice it will be used to adjust polling mode timer interval.
This trick dramatically reduces CPU usage in polling mode.


lvs1974:
@kprinssu, will you update submodule version in VoodooI2C before the next planned release?

kprinssu:
I’ll update the modules later today.

PR #27, opened 2020-04-03
Fix laggy trackpad after polling mode optimisation commit

lvs1974:
This is a fix for https://github.com/alexandred/VoodooI2C/issues/269.

### DIFF
--- a/VoodooI2CHID/VoodooI2CHIDDevice.cpp
+++ b/VoodooI2CHID/VoodooI2CHIDDevice.cpp
@@ -14,7 +14,7 @@
 #define super IOHIDDevice
 OSDefineMetaClassAndStructors(VoodooI2CHIDDevice, IOHIDDevice);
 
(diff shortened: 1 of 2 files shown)


kprinssu:
I'll sync changes to VoodooI2C master once we get a fix for https://github.com/alexandred/VoodooI2C/issues/269#issuecomment-608402938

lvs1974:
@kprinssu, my fix in VoodooI2CHID has nothing to do with issue which can be observed by @Mateo1234454545. It is something what we are trying to find out.

kprinssu:
@lvs1974 looks like removing the IOWorkLoop from @erictoby was the cause. Let's carry the discussion on there.

lvs1974:
@kprinssu, I think we do not need to fix anything else here.
You can read discussion in 269, and there were issues since release 2.1.

PR #28, opened 2020-04-10
Handle GPIO level interrupts properly

Goshin:
- Perform blocking read in ISR to reduce massive interrupts.
- Register direct interrupt for VoodooGPIO to block level interrupt while reading.

Check out some details here. https://github.com/alexandred/VoodooI2C/issues/259#issuecomment-611935065


Goshin:
This might fix some issues related to the GPIO mode.

ben9923:
Thanks @Goshin :)

PR #29, opened 2020-04-10
Fix some memory leaks/null dereferences

ben9923:


VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -73,23 +73,54 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
-    property_array->setObject("OutputRegister", OSNumber::withNumber(hid_descriptor.wOutputRegister, 32));
-    property_array->setObject("CommandRegister", OSNumber::withNumber(hid_descriptor.wCommandRegister, 32));
-    property_array->setObject("DataRegister", OSNumber::withNumber(hid_descriptor.wDataRegister, 32));
-    property_array->setObject("VendorID", OSNumber::withNumber(hid_descriptor.wVendorID, 32));
-    property_array->setObject("ProductID", OSNumber::withNumber(hid_descriptor.wProductID, 32));
-    property_array->setObject("VersionID", OSNumber::withNumber(hid_descriptor.wVersionID, 32));
+    if (!property_array)
+        return kIOReturnNoMemory;
+
+    OSNumber* hid_desc_length      = OSNumber::withNumber(hid_descriptor.wHIDDescLength,      32);
```
kprinssu, line 79: This can we be cleaned up with a small function that will take the number, property array, then create the OSNumber, set the property, then finally release the OSNumber.

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -722,23 +722,24 @@ inline void VoodooI2CMultitouchHIDEventDriver::setButtonState(DigitiserTransduce
...
-    if (!digitiser.transducers)
-        goto exit;
+    OSNumber* transducer_count = OSNumber::withNumber(digitiser.transducers->getCount(), 32);
 
-    properties->setObject("Contact Count Element", digitiser.contact_count);
-    properties->setObject("Input Mode Element", digitiser.input_mode);
-    properties->setObject("Contact Count Maximum  Element", digitiser.contact_count_maximum);
-    properties->setObject("Button Element", digitiser.button);
-    properties->setObject("Transducer Count", OSNumber::withNumber(digitiser.transducers->getCount(), 32));
+    properties->setObject("Contact Count Element",         digitiser.contact_count);
```
kprinssu, line 734: Don't you need to convert the values to an OSNumber?

ben9923: Everything but the transducer count is `IOHIDElement*`

PR #19, opened 2019-09-23
Mutex for smooth cursor movement.

Erictoby:
Added mutex to synchronize tasks, prevent recurring interrupt, fixes data corruption by recurring calls of read/write I2C, fixes precision lost from int to float conversion, fix several bugs.

PR #30, opened 2020-04-14
Implementation of @erictoby's "mutex for smooth cursor movement" fixing merge conflicts

marianopela:
I've been wanting to give https://github.com/alexandred/VoodooI2CHID/pull/19 a try for some time, and here it is

Could anyone test it and review it? I'm running fine in polling mode but don't have the possibility to test it with interrupts. Thanks

EDIT: This will hopefully fix https://github.com/alexandred/VoodooI2C/issues/284 and https://github.com/alexandred/VoodooI2C/issues/282
It fixed my sleep/wake problems. Here [voodoo log.txt](https://github.com/alexandred/VoodooI2CHID/files/4480743/voodoo.log.txt) was the problem.

It also, on my device, results in https://github.com/alexandred/VoodooI2C/pull/276 working as expected, as opposed to it not working before.


Goshin:
Does this fix solve a specific bug?

If it works as expected, the reading in VoodooI2CHID is in serial with the workloop. So I think we don't need to lock the `api->writeI2C` process.
In further, there is a state check for the I2C bus before starting transmission to prevent race.

.gitignore
```diff
@@ -20,4 +21,4 @@ xcuserdata/
 ## Other
 *.moved-aside
 *.xccheckout
-*.xcscmblueprint
\ No newline at end of file
+*.xcscmblueprint.DS_Store
```
hieplpvip, line 24: This doesn't seem right for me. Are you missing new line?

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
-    unsigned char* report = (unsigned char *)IOMalloc(hid_descriptor.wMaxInputLength);
-
-    api->readI2C(report, hid_descriptor.wMaxInputLength);
+    int return_size;
+    unsigned char* report;
     
-    int return_size = report[0] | report[1] << 8;
+    if (I2C_TRYLOCK() == false) {
+        IOLog("%s::%s Skipping a HID read interrupt while other thread read/write I2C\n", getName(), name);
+        goto exit;
```
hieplpvip, line 192: `goto exit;` essentially means assigning `read_in_progress = false;`, which is wrong since another thread is reading/writing I2C and mutex is locked by that thread, not current thread. This could lead to undefined behavior.

I think we should `return;` here.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
+    }
+    report = getMallocI2CIntr(hid_descriptor.wMaxInputLength);
+    report[0] = report[1] = 0;
+    ret = api->readI2C(report, hid_descriptor.wMaxInputLength);
+    if (ret != kIOReturnSuccess) {
+        I2C_UNLOCK();
```
hieplpvip, line 198: As I said above, `goto exit;` sets `read_in_progress = false;`. However, we unlock the mutex before assigning value. This again could lead to undefined behavior.

I think we should call `I2C_UNLOCK();` after `read_in_progress = false;` in `exit`

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
 
+    return_size = report[0] | report[1] << 8;
+    /*
+     * "return_size" can be 0 when resetHIDDevice() is called; booting or waking up from sleep.
+     * Since ready_for_input can still be FALSE when booting,
+     * It needs to be checked before checking ready_for_input.
+     */
     if (!return_size) {
-        // IOLog("%s::%s Device sent a 0-length report\n", getName(), name);
+        I2C_UNLOCK();
```
hieplpvip, line 209: Same as above

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
         goto exit;
     }
 
-    if (!ready_for_input)
-        goto exit;
-
-    if (return_size > hid_descriptor.wMaxInputLength) {
-        // IOLog("%s: Incomplete report %d/%d\n", getName(), hid_descriptor.wMaxInputLength, return_size);
+    if (!ready_for_input || return_size > hid_descriptor.wMaxInputLength) {
+        I2C_UNLOCK();
```
hieplpvip, line 215: Same as above

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
         goto exit;
     }
 
     buffer = IOBufferMemoryDescriptor::inTaskWithOptions(kernel_task, 0, return_size);
     buffer->writeBytes(0, report + 2, return_size - 2);
     
+    I2C_UNLOCK();
```
hieplpvip, line 222: Should delete this line as `I2C_UNLOCK()` will be called later in `exit`

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -148,36 +184,49 @@ IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
...
     ret = handleReport(buffer, kIOHIDReportTypeInput);
 
     if (ret != kIOReturnSuccess)
         IOLog("%s::%s Error handling input report: 0x%.8x\n", getName(), name, ret);
     
     buffer->release();
-    IOFree(report, hid_descriptor.wMaxInputLength);
 
 exit:
     read_in_progress = false;
```
hieplpvip, line 232: Call `I2C_UNLOCK()` after this line

marianopela:
> Does this fix solve a specific bug?

Not yet. The problem is my trackpad not working after some sleep/wake. Still present, unfortunately. I wanted to present those changes as they were when @erictoby originally submitted the PR back in November and see what you guys think about it. There is also https://github.com/alexandred/VoodooI2C/issues/284 which is strange

hieplpvip:
Hi, I think there are some problems with the current implementation. I've commented on them. Could anyone review it?

hieplpvip:
This pull request breaks single click for me. Single click becomes triple click, but FingerMgmt only shows 1 finger.

marianopela:
> This pull request breaks single click for me. Single click becomes triple click, but FingerMgmt only shows 1 finger.

Ok this I can address immediately. It happens when Force Touch is enabled. Isn't it what force touch is all about? 
Screenshot 2020-04-14 at 20 42 30



Single click shows apps in Launchpad, renames files and shows lookup. But I might be mistaken and wanted to ask @Goshin 's opinion on this, in relation to https://github.com/alexandred/VoodooI2C/pull/276

I now feel like it didn't work properly before

hieplpvip:
> > This pull request breaks single click for me. Single click becomes triple click, but FingerMgmt only shows 1 finger.
> 
> Ok this I can address immediately. It happens when Force Touch is enabled. Isn't it what force touch is all about?
> 
> Single click shows apps in Launchpad, renames files and shows lookup. But I might be mistaken and wanted to ask @Goshin 's opinion on this, in relation to [alexandred/VoodooI2C#276](https://github.com/alexandred/VoodooI2C/pull/276)
> 
> I now feel like it didn't work properly before (it worked with 3 fingers tap only)

Thank you. I didn't know Force Touch simulation was added recently

marianopela:
> Thank you. I didn't know Force Touch simulation was added recently

Take what I said as my theory, until someone more authoritative can confirm or deny it.
Anyhow, do you think some of those changes, with all due adjustments, can result in a better implementation of the HID satellite?

I started looking into these because my trackpad isn't working properly after sleep and this happens
[log.txt](https://github.com/alexandred/VoodooI2CHID/files/4477809/log.txt)

hieplpvip:
My trackpad doesn't stop working after sleep, but it randomly stops, and the only way to fix it is to sleep and wake. I guess there's a race condition related to `read_in_progress`. These changes may fix it.

marianopela:
So it's kind of a common problem. I confirm putting it to sleep and waking fixes it but not always at first attempt. I tried building with your changes but unfortunately it results in a kernel panic at boot

hieplpvip:
@marianopela I've fixed it. Forgot to remove one `I2C_UNLOCK()` call.

marianopela:
> @marianopela I've fixed it. Forgot to remove one `I2C_UNLOCK()` call.

It booted, thanks. I'm gonna test it for a while, with some sleep/wake and report back. Hope your random stops are fixed too

marianopela:
@hieplpvip In the meantime, are you running in interrupt mode? Because I still have no proof that this doesn't break interrupt mode, as I run in polling only

hieplpvip:
Interrupt doesn't work for me. Need someone else to test

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -369,14 +371,14 @@ IOReturn VoodooI2CHIDDevice::resetHIDDevice() {
 IOReturn VoodooI2CHIDDevice::resetHIDDeviceGated() {
     setHIDPowerState(kVoodooI2CStateOn);
 
+    read_in_progress = true;
```
hieplpvip, line 373: This should be after `I2C_LOCK();`

hieplpvip:
@marianopela You made a small mistake. Otherwise, everything is good.

marianopela:
> @marianopela You made a small mistake. Otherwise, everything is good.

You're right. I was kind of in a rush, thanks.

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -270,12 +280,16 @@ bool VoodooI2CTouchscreenHIDEventDriver::handleStart(IOService* provider) {
...
+        timer_source->cancelTimeout();
         work_loop->removeEventSource(timer_source);
         OSSafeReleaseNULL(timer_source);
     }
-    
-    OSSafeReleaseNULL(work_loop);
-    
+
+    if (work_loop) {
+        // OSSafeReleaseNULL(work_loop);
```
kprinssu, line 289: You definetly want to release `work_loop`. In fact, `OSSafeReleaseNULL(work_loop);` is perfectly fine here.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -322,48 +369,58 @@ IOReturn VoodooI2CHIDDevice::resetHIDDevice() {
 
 IOReturn VoodooI2CHIDDevice::resetHIDDeviceGated() {
     setHIDPowerState(kVoodooI2CStateOn);
+
     read_in_progress = true;
+    I2C_LOCK();
+    VoodooI2CHIDDeviceCommand* command = (VoodooI2CHIDDeviceCommand*) getMallocI2C(sizeof(VoodooI2CHIDDeviceCommand));
```
kprinssu, line 375: Please use stack allocations as you're leaking memory.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -38,27 +39,65 @@ bool VoodooI2CHIDDevice::init(OSDictionary* properties) {
...
+
+    return retaddr;
+}
+
 IOReturn VoodooI2CHIDDevice::getHIDDescriptor() {
-    VoodooI2CHIDDeviceCommand command;
-    command.c.reg = hid_descriptor_register;
+    I2C_LOCK();
+    read_in_progress = true;
+    VoodooI2CHIDDeviceCommand* command = (VoodooI2CHIDDeviceCommand*)getMallocI2C(sizeof(VoodooI2CHIDDeviceCommand));
```
kprinssu, line 86: Please use stack allocations.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -38,27 +39,65 @@ bool VoodooI2CHIDDevice::init(OSDictionary* properties) {
...
+    IOFree(buf_i2c_pool, I2C_MAX_BUF_SIZE);
+    if (read_in_progress_mutex) {
+        IOLockFree(read_in_progress_mutex);
+        read_in_progress_mutex = NULL;
+    }
 
     super::free();
 }
 
+UInt8* VoodooI2CHIDDevice::getMallocI2CIntr(UInt16 size) {
```
kprinssu, line 65: What happens when the buffer is full? It seems like you overwrite memory.

Can you explain to me why we even need a buffer?

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -38,27 +39,65 @@ bool VoodooI2CHIDDevice::init(OSDictionary* properties) {
...
 }
 
 void VoodooI2CHIDDevice::free() {
     if (client_lock)
         IOLockFree(client_lock);
+    
+    IOFree(buf_i2c_pool_intr, I2C_MAX_BUF_SIZE);
+    IOFree(buf_i2c_pool, I2C_MAX_BUF_SIZE);
+    if (read_in_progress_mutex) {
+        IOLockFree(read_in_progress_mutex);
```
kprinssu, line 58: How do you `read_in_progress_mutex` is not NULL?

.gitignore
```diff
@@ -20,4 +20,5 @@ xcuserdata/
 ## Other
 *.moved-aside
 *.xccheckout
-*.xcscmblueprint
\ No newline at end of file
+*.xcscmblueprint.
+DS_Store
```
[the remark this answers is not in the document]
hieplpvip: `.gitignore` doesn't include `.DS_Store` yet

hieplpvip:
The idea of using a buffer is introduced by @Erictoby. He commented on it:
```/* Buffers for <api->readI2C>, <api->writeI2C>, <api->writeReadI2C>
     *
     * Since the functions may use the buffers even after the end of calls,
     * the buffers need to be kept after calls.
     */
```
I'm not sure if it's really necessary, we'd better ask him. In the meantime, I suggest reverting the buffer change, fixing memory allocation and merging this pull request.

@kprinssu Is that okay with you?

kprinssu:
@hieplpvip I understand why the buffers exist, but why not use stack allocations? It's much much faster and you don't need to keep track of memory. @erictoby probably realised the same thing but in my opinion, implemented it in the wrong way.

kprinssu:
@marianopela your wake issue is probably fixed with this line, https://github.com/alexandred/VoodooI2CHID/pull/30/files#diff-879e63994a863a726fb58e4c54bd1fcaR519. Moving `awake = true` after the I2C command; I am assuming interrupts start happening and your hardware gets into a bad state since it's still hasn't gotten the "wake up" commands.

Please try moving that bit of code and see if that fixes your wake up issue.

marianopela:
> Moving `awake = true` after the I2C command
> Please try moving that bit of code and see if that fixes your wake up issue.

Do you mean after the unlock?

Anyway, @Goshin solved it [here](https://github.com/alexandred/VoodooI2C/issues/284#issuecomment-614475630)

I think I can delete this PR, in favour of https://github.com/alexandred/VoodooI2CHID/pull/31 which is much better.

PR #32, opened 2020-04-26
Fix sleep issues

Goshin:
- Disable interrupt timer before entering sleep. Fix power state transition timeout in https://github.com/alexandred/VoodooI2C/issues/284.
- Remove `read_in_progress` to avoid the potential race condition. The lock is moved to I2C controller side.
- Disable GPIO interrupt before entering sleep.


kprinssu:
@ben9923 what do you think?

PR #33, opened 2020-04-30
Extract interrupt toggling to functions

Goshin:
Extract interrupt toggling to functions and unregister interrupt when entering the sleep state, similar to https://github.com/alexandred/VoodooI2C/pull/298.


kprinssu:
Thanks @Goshin. Sorry for the delay, this looks good and mostly a refactor.

PR #35, opened 2020-05-15
Fix bad argument in waiting for device reset

Goshin:
- Remove `IOSleep(100)`. `api->writeI2C()` is blocking and synchronized, so the `IOSleep` here is not neccesary.
- Fix the bad argument for `command_gate->commandSleep`. The second argument of `command_gate->commandSleep` is a clock deadline. Currently, passing a time period as the argument causes an immediate wake-up.
- Add a log for the successful reset.


kprinssu:
Good catch as usual @Goshin. IOCommandGate expects a `AbsoluteTimedeadline` and not a AbsoluteTime.

PR #37, opened 2020-08-07
Refactor getHIDDescriptorAddress(); fix transport string comparison.

zhen-zen:


kprinssu:
@zhen-zen this looks good, but do you providing context behind this PR?

zhen-zen:
Sure. Related document is https://docs.microsoft.com/en-us/windows-hardware/drivers/bringup/hidi2c-device-specific-method---dsm- . What I changed is to read the guid using `uuid_parse` from a human readable format instead. And since the `_DSM` method will only returns an integer when that guid is provided, `evaluateInteger` is much simpler for processing.

kprinssu:
This looks good to me, @ben9923 can you take a second look?

zhen-zen:
Before merging, I wonder if we can move ACPI evaluation to VoodooI2C. I'm working on another PR on _DSM there.

zhen-zen:
Related PR is alive at https://github.com/VoodooI2C/VoodooI2C/pull/365

kprinssu:
@zhen-zen I do not agree that it should be in VoodooI2C. HID kext pertains everything to HID and here we are reporting the descriptor to the IO Registry. In terms scope, I think it should stay here rather within VoodooI2C.

zhen-zen:
@kprinssu I suppose the HIDDescriptor could act as an indicator of whether the I2C device support HID. I did some quick search on DSDTs of specific I2C satellite models. Generally, the address just return 0 for non-HID device. Correct me if I missed some cases.

Another exception is VoodooI2CSynaptics. This driver is actually built on top of hid-rmi and rmi-i2c layers. The hardcoded values in I2C packets are actually HID registers.

This move could remove the usage of `IOACPIPlatformDevice` in satellite drivers and simplify the workaround of adding ACPI device to IORegistry.

Personally, I also purposed this change due to the weird KP when casting `acpi-device` to `IOACPIPlatformDevice` in https://github.com/VoodooSMBus/VoodooRMI/pull/42.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -99,19 +99,19 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
 IOReturn VoodooI2CHIDDevice::getHIDDescriptorAddress() {
     uuid_t guid;
-    uuid_parse(HID_I2C_DSM_HIDG, guid);
+    uuid_parse(I2C_DSM_HIDG, guid);
 
     // convert to mixed-endian
-    *(uint32_t *)guid = OSSwapInt32(*(uint32_t *)guid);
-    *((uint16_t *)guid + 2) = OSSwapInt16(*((uint16_t *)guid + 2));
-    *((uint16_t *)guid + 3) = OSSwapInt16(*((uint16_t *)guid + 3));
+    *(reinterpret_cast<uint32_t *>(guid)) = OSSwapInt32(*(reinterpret_cast<uint32_t *>(guid)));
```
kprinssu, line 105: Please avoid using `reinterpret_cast` or any of the other C++ casts. 

The mach kernel has it's own type definition table and it's highly recommended to use `OSDynamicCast` instead.

See documentation here:
https://developer.apple.com/documentation/driverkit/osdynamiccast

zhen-zen: I changed that since VoodooI2C main project prompts so at linting stage. However, in libkern/libkern/c++/OSObject.h, it says "OSDynamicCast is build on the OSMetaClass mechanism". Or if the driverkit's version got some enhancement?

kprinssu: You are correct. This has nothing to do with OS class definitions as OSSwapInt32 converts kernel specific numbers into integers. Please ignore the previous comment.

kprinssu:
@zhen-zen It looks good to me as-is. Are you done with changes on your end?

zhen-zen:
Some minor changes, that's all.

According to ACPI specs, the availability of specific function index (HID register) for _DSM should be checked first. However, each I2C HID device should have this function and I suppose it's safe to skip that step.

And it seems that the "ACPI Warning: \_SB_.PCI0.I2C1.TPL0._DSM: Argument #4 type mismatch - Found [Integer], ACPI requires [Package]" is also resolved by replacing Arg3 to `OSArray::withCapacity(1)`.

zhen-zen:
Wait a minute, maybe that arg could be `OSArray::withCapacity(0)`?

Goshin:
Maybe you can reuse `VoodooI2CDeviceNub::evaluateDSM()` in https://github.com/VoodooI2C/VoodooI2C/pull/365?

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -374,7 +374,7 @@ bool VoodooI2CMultitouchHIDEventDriver::handleStart(IOService* provider) {
     if (!transport)
         return false;
 
-    if (transport->getCStringNoCopy() != kIOHIDTransportUSBValue)
+    if (strncmp(transport->getCStringNoCopy(), kIOHIDTransportUSBValue, sizeof(kIOHIDTransportUSBValue)))
```
ben9923, line 377: Good catch!

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -98,53 +98,35 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
+
+    UInt32 result;
+    OSObject *params[4] = {
+        OSData::withBytes(guid, 16),
+        OSNumber::withNumber(I2C_DSM_REVISION, 8),
+        OSNumber::withNumber(HIDG_DESC_INDEX, 8),
+        OSArray::withCapacity(1)
+    };
+
+    if (acpi_device->evaluateInteger("_DSM", &result, params, 4) != kIOReturnSuccess && acpi_device->evaluateInteger("XDSM", &result, params, 4) != kIOReturnSuccess) {
```
ben9923, line 117: Like the usage of `evaluateInteger`.
I still do prefer to split the `_DSM` and `XDSM` evaluation into separate conditions, as it's just too long to read..

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -98,53 +98,35 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
-    if (number) {
-        setProperty("HIDDescriptorAddress", number);
-        hid_descriptor_register = number->unsigned16BitValue();
-    }
 
-    if (result)
-        result->release();
-    
+    setProperty("HIDDescriptorAddress", result, 32);
+    hid_descriptor_register = (UInt16) result;
```
ben9923, line 123: Please avoid C-style casts.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -98,53 +98,35 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
+    uuid_t guid;
+    uuid_parse(I2C_DSM_HIDG, guid);
+
+    // convert to mixed-endian
+    *(reinterpret_cast<uint32_t *>(guid)) = OSSwapInt32(*(reinterpret_cast<uint32_t *>(guid)));
+    *(reinterpret_cast<uint16_t *>(guid) + 2) = OSSwapInt16(*(reinterpret_cast<uint16_t *>(guid) + 2));
+    *(reinterpret_cast<uint16_t *>(guid) + 3) = OSSwapInt16(*(reinterpret_cast<uint16_t *>(guid) + 3));
+
+    UInt32 result;
+    OSObject *params[4] = {
```
ben9923, line 110: Size (`4`) is unnecessary here.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -98,53 +98,35 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
+        OSData::withBytes(guid, 16),
+        OSNumber::withNumber(I2C_DSM_REVISION, 8),
+        OSNumber::withNumber(HIDG_DESC_INDEX, 8),
+        OSArray::withCapacity(1)
```
ben9923, line 114: A trailing comma is always encouraged after every line :)

VoodooI2CHID/VoodooI2CHIDDevice.cpp
zhen-zen: Maybe just a new line?

zhen-zen:
> Maybe you can reuse `VoodooI2CDeviceNub::evaluateDSM()` in [VoodooI2C/VoodooI2C#365](https://github.com/VoodooI2C/VoodooI2C/pull/365)?

The method used here is a simplified version for HID descriptor case. Expected return is an integer and there's no need to test if the function index is available.

Goshin:
> > Maybe you can reuse `VoodooI2CDeviceNub::evaluateDSM()` in [VoodooI2C/VoodooI2C#365](https://github.com/VoodooI2C/VoodooI2C/pull/365)?
> 
> The method used here is a simplified version for HID descriptor case. Expected return is an integer and there's no need to test if the function index is available.

Yes but it seems `VoodooI2CDeviceNub::evaluateDSM()` doesn't include the validation of the function index? The only difference I can see is the return type, so the extra lines needed here will be casting the result type, but we can greatly simplify the manipulation of GUID and other parameters if reuse the function.

gongzhenjeef:
> > > Maybe you can reuse `VoodooI2CDeviceNub::evaluateDSM()` in [VoodooI2C/VoodooI2C#365](https://github.com/VoodooI2C/VoodooI2C/pull/365)?
> > 
> > The method used here is a simplified version for HID descriptor case. Expected return is an integer and there's no need to test if the function index is available.
> 
> Yes but it seems `VoodooI2CDeviceNub::evaluateDSM()` doesn't include the validation of the function index? The only difference I can see is the return type, so the extra lines needed here will be casting the result type, but we can greatly simplify the manipulation of GUID and other parameters if reuse the function.

VoodooI2C has polling mode for fallback and the usage of TP7G is undocumented, so I followed the ACPI spec given available information. The implementation there allows future expansion if new information is given. 

However, HID descriptor is critical for VoodooI2CHID and HIDG is well-documented officially. So I suppose it's safe to skip the long implementation.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -98,53 +98,25 @@ IOReturn VoodooI2CHIDDevice::parseHIDDescriptor() {
...
-    
-    acpi_device->evaluateObject("_DSM", &result, params, 4);
-    if (!result)
-        acpi_device->evaluateObject("XDSM", &result, params, 4);
-    if (!result) {
-        IOLog("%s::%s Could not find suitable _DSM or XDSM method in ACPI tables\n", getName(), name);
+    OSObject* result = nullptr;
+    if (api->evaluateDSM(I2C_DSM_HIDG, HIDG_DESC_INDEX, &result) != kIOReturnSuccess) {
+        IOLog("%s::%s unable to parse HID descriptor address\n", getName(), name);
+        result->release();
```
Goshin, line 104: `result` might be null if `api->evaluateDSM()` fails.

zhen-zen: Almost forgot that macro. That's why I don't like evaluateObject.

zhen-zen:
I moved the previous implementation to VoodooI2CSensorHubEnabler, which has no access to VoodooI2CDeviceNub but also need to evaluate `_DSM`. However, there's lack of information on that method and supplied GUID. A sample of that SSDT is attached below.
[SSDT-1-sensrhub.aml.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/5081807/SSDT-1-sensrhub.aml.zip)

kprinssu (approved):

ben9923 (approved):

kprinssu:
This PR is good to go. However, we need to merge in https://github.com/VoodooI2C/VoodooI2C/pull/365/files before we can go ahead.

PR #38, opened 2020-10-01
Fix potential null dereference and string comparison

hieplpvip:


kprinssu:
Looks good to me. Thanks @hieplpvip 

If @ben9923 okay's it, then it's good for a merge :).

ben9923 (approved):
Looks good, thanks!

PR #43, opened 2021-01-03
Don't add/remove event source on sleep/wake

ben9923:
Adding/removing an interrupt event source from the work loop on sleep/wake makes it, according to Darwin-XNU resources, unregister & register the interrupt on those cases.

New VoodooGPIO functionally for fixing post-sleep pin IRQ issues on some machines requires pins to be registered on sleep to know which pins should be restored.
This change makes VoodooI2CHID only enable/disable the event source, without adding/removing it.

Other VoodooI2C satellites only add/remove the event source on `start/stop`, as they should.

Partially solves VoodooI2C/VoodooI2C#388


kprinssu:
Also looks good!

PR #44, opened 2021-01-09
Skip reset on wake up for 0911:5288 and 2386:3118

zhen-zen:
Implementation of `I2C_HID_QUIRK_NO_IRQ_AFTER_RESET`. https://patchwork.kernel.org/project/linux-input/patch/20171106150041.13933-1-hdegoede@redhat.com/
Fix https://github.com/VoodooI2C/VoodooI2C/issues/399

### DIFF
--- a/VoodooI2CHID/VoodooI2CHIDDevice.cpp
+++ b/VoodooI2CHID/VoodooI2CHIDDevice.cpp
@@ -30,6 +30,8 @@ bool VoodooI2CHIDDevice::init(OSDictionary* properties) {
     interrupt_source = NULL;
     ready_for_input = false;
 
+    quirks = 0;
+
     client_lock = IOLockAlloc();
 
     clients = OSArray::withCapacity(1);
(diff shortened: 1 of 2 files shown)


ben9923:
Thanks! Though it's already implemented here  :/
https://github.com/VoodooI2C/VoodooI2CHID/tree/device-quirks
Just waiting for confirmation from @m0d16l14n1

It's a more scalable way for future quirk implementations, too :P

Note you were mixing up two quirks. `I2C_HID_QUIRK_NO_IRQ_AFTER_RESET` means the device won't send an IRQ after reset, `I2C_HID_QUIRK_RESET_ON_RESUME` means we simply won't reset the device on wake, just power it on.
We currently (`master` code) reset all devices, apparently we shouldn't do so, besides two certain devices.

zhen-zen:
So reset after wake is not required on most PTP HID devices. I was confused since it is necessary for RMI mode and mixed them.

Also maybe it's `quirks |= i2c_hid_quirks[n].quirks;` here? https://github.com/VoodooI2C/VoodooI2CHID/blob/d250c1d7c4931e2f30baf99452004e7923b902c0/VoodooI2CHID/VoodooI2CHIDDevice.cpp#L407

ben9923:
@zhen-zen Linux seems to only get the last match for its quirks. Probably to be sorted from generic (i.e. every ELAN) to specific (i.e. ELAN1200) when needed.

m0d16l14n1:
@zhen-zen @ben9923 

Yep, it's seems to be working good. Tested with some short sleep-wake cycles and one "long". 

(Actually I am testing that already for 5 hours, didn't get any problem). 

That's the first time i get my touchpad working after sleep in every wake up. Previously i mostly get it working 1 of 50 or more attempts. 

Thanks to all the developers, who helped me in that journey to fix that problem.

PR #45, opened 2021-01-09
Add device quirks support + reset-related quirks

ben9923:
Besides the addition of quirks, devices are no longer being reset after wake - It seems to cause issues for various devices.
Some specific devices (currently 2) do need the reset, so they were added as a quirk.
Quirks come from the Linux i2c-hid driver.

Should resolve more sleep issues, including VoodooI2C/VoodooI2C#399


m0d16l14n1:
https://github.com/VoodooI2C/VoodooI2CHID/pull/44#issuecomment-757084470

kprinssu:
Looks good @ben9923

kprinssu (approved):

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -411,15 +425,11 @@ IOReturn VoodooI2CHIDDevice::setPowerState(unsigned long whichState, IOService*
...
-            command.c.opcode = 0x01;
-            command.c.report_type_id = 0;
-
-            api->writeI2C(command.data, 4);
-            IOSleep(10);
+            if (quirks & I2C_HID_QUIRK_RESET_ON_RESUME) {
+                resetHIDDevice();
+            } else {
+                setHIDPowerState(kVoodooI2CStateOn);
+            }
```
kprinssu, line 432: Is it safe to remove the `IOSleep` here?

ben9923: `setHIDPowerState` sleeps for `100ms` after setting it.
`resetHIDDevice` calls `setHIDPowerState` inside, then either waits for an IRQ (on almost all devices) or waits `100ms` after the reset.

kprinssu: Makes sense, go for the merge!

PR #46, opened 2021-01-11
Publish "RM,deliverNotifications" in handleStart for possible race condition

zhen-zen:
https://github.com/VoodooI2C/VoodooI2C/issues/427
https://github.com/acidanthera/bugtracker/issues/1415

CC @witampanstwa @lvs1974 @ben9923

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -100,8 +100,6 @@
 			<string>VoodooI2CPrecisionTouchpadHIDEventDriver</string>
 			<key>IOProviderClass</key>
 			<string>IOHIDInterface</string>
-			<key>RM,deliverNotifications</key>
-			<true/>
 			<key>QuietTimeAfterTyping</key>
 			<integer>100</integer>
 			<key>ProcessUSBMouseStopsTrackpad</key>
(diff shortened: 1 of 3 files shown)


ben9923:
From my understanding the issue is fixed with just https://github.com/acidanthera/VoodooPS2/commit/dc3571249f59d3aa0bfd867596bb7121373c8ee9, isn't it?

lvs1974:
@ben9923: at least that fix solved issue for me and @witampanstwa, so I think it should work for everyone.

witampanstwa:
> @ben9923: at least that fix solved issue for me and @witampanstwa, so I think it should work for everyone.

That is, if your configuration uses GPIO interrupts because mine does use pooling (unless you're sure that it does indeed work in the GPIO mode too)

zhen-zen:
I think the workaround is weird for a race condition issue but @lvs1974 discovered that the `IONotify` pointer could differ. Later I dug into the IOKit source code [further](https://github.com/acidanthera/bugtracker/issues/1415#issuecomment-758343804) but it seems that the mechanism should work reliably. Since I don't have resources to test it right now, I will just close this PR.

PR #48, opened 2021-01-26
Enable USB precision touchpad support

Kethen:
On bigsur 11.1 along with the current master branch of VoodooI2C, `digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);` works on both I2C(lenovo yoga530) and USB(keyboard dock for a acer switch 11v sw5-173), with one quirk being the 3 points touch usb touchpad does not show up in system preference->trackpad and I have yet to figure out why (after some reading probably due to the unpatched broadwell acpi battery not working)

existing implementation utilizing `hid_interface->setReport` does not work on USB; touchpad does not enter touchpad mode hence sending 0 coordinates only, hotplugging the keyboard dock causing the whole dock to stop functioning, while `digitiser.input_mode->setValue` works on both my I2C and USB touchpads

### DIFF
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
@@ -12,9 +12,6 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
-    // We should really do this using `input_mode_element->setValue(INPUT_MODE_TOUCHPAD)`
-    // but I am not able to get it to work.
-
     digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);
 
     ready = true;
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
@@ -15,15 +15,7 @@ void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
     // We should really do this using `input_mode_element->setValue(INPUT_MODE_TOUCHPAD)`
     // but I am not able to get it to work.
 
-    VoodooI2CPrecisionTouchpadFeatureReport buffer;
-    buffer.value = INPUT_MODE_TOUCHPAD;
-    buffer.reserved = 0x00;
-
-    IOBufferMemoryDescriptor* report = IOBufferMemoryDescriptor::inTaskWithOptions(kernel_task, 0, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
-    report->writeBytes(0, &buffer, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
-
-    hid_interface->setReport(report, kIOHIDReportTypeFeature, digitiser.input_mode->getReportID());
-    report->release();
+    digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);
 
     ready = true;
 }
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -95,11 +95,6 @@
 			</array>
 			<key>IOMatchCategory</key>
 			<string>VoodooI2CHIDMultitouch</string>
-			<key>IOPropertyMatch</key>
-			<dict>
-				<key>Transport</key>
-				<string>I2C</string>
-			</dict>
 			<key>IOProbeScore</key>
 			<integer>300</integer>
 			<key>IOClass</key>
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
@@ -23,11 +23,6 @@
 #define INPUT_MODE_MOUSE 0x00
 #define INPUT_MODE_TOUCHPAD 0x03
 
-typedef struct __attribute__((__packed__)) {
-    UInt8 value;
-    UInt8 reserved;
-} VoodooI2CPrecisionTouchpadFeatureReport;
-
 /* Implements an HID Event Driver for Precision Touchpad devices as specified by Microsoft's protocol in the following document: https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/precision-touchpad-devices
  *
  * The members of this class are responsible for instructing a Precision Touchpad device to enter Touchpad mode.


jlempen:
Hi @Kethen,
Thank you so much for this! 
I can confirm that your commit works perfectly for the Synaptics USB trackpad found in the detachable keyboard dock of my **Lenovo ThinkPad X1 Tablet Gen 3**, as well as for the USB trackpad found in the detachable type cover of my **Microsoft Surface Go 2** (the highest end model with the m3-8100Y processor). I'm still on Catalina 10.5.7 on both tablets. Without your commit, both trackpads won't work at all with VoodooI2C & the VoodooI2CHID satellite and work as a simple mouse if I remove the VoodooI2C kexts entirely (or use Karabiner Elements to handle them). VoodooI2C actually detects the precision trackpads and seems to activate them, but they won't work at all.
As I don't have any device with an I2C trackpad on hand, I'm afraid I won't be able to help you check if the commit won't break other I2C trackpads, though.
I confirm as well that the I2C touchscreens handled by VoodooI2CHID keep working on both tablets with your commit. The touchscreen is a Synaptics on the Lenovo and an ELAN on the Microsoft tablet.
I have working battery status on both tablets and therefore the trackpad settings appear in system preferences. They appear to be shared with the touchscreen, though. That's the same issue as when you use a Magic Trackpad or Magic Mouse on a MacBook. The internal trackpad and the Magic Trackpad share the same settings.
Please tell me if you need more information, logs or anything to help you getting this commit merged.

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -15,15 +15,7 @@ void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
     // We should really do this using `input_mode_element->setValue(INPUT_MODE_TOUCHPAD)`
     // but I am not able to get it to work.
```
ben9923, line 16: You probably want to remove those too... :P

ben9923:
@Kethen Thanks!
We definitely need some more I2C precision trackpad testers for this one before releasing it, to make sure we don't break anything.

zhen-zen:
Congrats you found a better implementation.

Don't own my old tablet with USB PTP topcase any more. iirc if I just remove `Transport` restriction from `VoodooI2CPrecisionTouchpadHIDEventDriver` (or `VoodooI2CMultitouchHIDEventDriver` by default), the touchpad would enter PTP mode but won't respond to any touch events. I wonder if it's the same case as yours.

Without VoodooI2C (sometimes I need to add the same `Transport` match to prevent `VoodooI2CMultitouchHIDEventDriver` from attaching), it will just behave like a single mouse as @jlempen said above.

Just gave it a try on a separate USB PTP touchpad bought recently, but it still won't work. The good news is, `setReport` seems to work since the trackpad will no longer output fake key comb, i.e. three finger swipe converted to a series of `[C` stuffs. So it stops reporting in compatible mode, which is usual for such gadgets. However, I don't have much time to play with it right now to dig in further.

Xiashangning:
> Hi @Kethen,
> Thank you so much for this!
> I can confirm that your commit works perfectly for the Synaptics USB trackpad found in the detachable keyboard dock of my **Lenovo ThinkPad X1 Tablet Gen 3**, as well as for the USB trackpad found in the detachable type cover of my **Microsoft Surface Go 2** (the highest end model with the m3-8100Y processor). I'm still on Catalina 10.5.7 on both tablets. Without your commit, both trackpads won't work at all with VoodooI2C & the VoodooI2CHID satellite and work as a simple mouse if I remove the VoodooI2C kexts entirely (or use Karabiner Elements to handle them). VoodooI2C actually detects the precision trackpads and seems to activate them, but they won't work at all.
> As I don't have any device with an I2C trackpad on hand, I'm afraid I won't be able to help you check if the commit won't break other I2C trackpads, though.
> I confirm as well that the I2C touchscreens handled by VoodooI2CHID keep working on both tablets with your commit. The touchscreen is a Synaptics on the Lenovo and an ELAN on the Microsoft tablet.
> I have working battery status on both tablets and therefore the trackpad settings appear in system preferences. They appear to be shared with the touchscreen, though. That's the same issue as when you use a Magic Trackpad or Magic Mouse on a MacBook. The internal trackpad and the Magic Trackpad share the same settings.
> Please tell me if you need more information, logs or anything to help you getting this commit merged.

Hi, can your Surface Go 2's keyboard work? I have made a PR for SP 7's keyboard. With the official version, only the trackpad get recognized and all gestures are usable. Now with my patch, the keyboard works too.
Besides, I also have a question. How does your Surface Go 2's touch screen as well as the battery works. None of them works on my SP 7 (I think that these hardware specs are similar). A dumped DSDT or some custom patches will be insanely helpful. I would really like to make these two components work on my SP7. Thanks.

lulujyc:
Hello @Kethen ,
I've tested this driver on the HP Elite X2 G4 trackpad, which is provided by ELAN instead of Synaptics (as for the case of X1 Tablet Evo). When the Multitouch HID Event Driver is being used, the trackpad does not work at all. I wonder if you wouldn't mind helping ELAN trackpad users. Another case showing the same symptom can be found here: https://github.com/acai66/lenovo-miix-520-hackintosh

jlempen:
> Hi, can your Surface Go 2's keyboard work? I have made a PR for SP 7's keyboard. With the official version, only the trackpad get recognized and all gestures are usable. Now with my patch, the keyboard works too.
> Besides, I also have a question. How does your Surface Go 2's touch screen as well as the battery works. None of them works on my SP 7 (I think that these hardware specs are similar). A dumped DSDT or some custom patches will be insanely helpful. I would really like to make these two components work on my SP7. Thanks.

Hi @Xiashangning,
Sorry for my very late reply!
I'm afraid the Surface Go 2 and the Surface Pro 7 are very different devices which have only the "Surface" name in common. The Go 2 doesn't need any DSDT patch for battery status and the touch screen works nicely with VoodooI2CHID. The keyboard works out of the box with VoodooPS2. Only the touchpad needs this PR to work. 
Unfortunately, the SP7's touchscreen uses IPTS and will most likely never work on macOS. The battery status should be fixable with DSDT patches.

kprinssu:
@Xiashangning @jlempen Please keep the discussion to VoodooI2C. Can you please take the discussion of Hackintosh off-here please?

fenglinnet:
@Xiashangning @Kethen Thanks！
I merged your PR. On my Surface Book 2, the keyboard and trackpad (gestures) drive almost perfectly, except that the trackpad can't be detected in the preferences.

kingo132:
Hi, thanks for this fix. My surface go 2's trackpad is now working. But the acceleration seems too fast, it's very hard to click an item, even a small move will make the cursor move across the entire screen. I set the trackpad speed to zero in system preference. Still the same. Do you have any clue about how to fix this trackpad acceleration or speed?

kingo132:
I think I found the reason, the physical_max is multiplied by 10 in VoodooI2CHID. This gives the result of 95.5x53.08cm size of the trackpad of surface type cover of surface go2, the actual size of it is 9.55x5.38cm.
image
This line should be removed. I don't know if it was only applied to surface go2. Or maybe it is a bug. I'll post this under that issue I mentioned.

kprinssu:
@kingo132 That change was added to fix scaling issues across other hardware. The driver (or this patch) should report units in `mm` not `cm`. Hence why you see the scaling by a factor of 10.

kingo132:
> Hi @Kethen,
> Thank you so much for this!
> I can confirm that your commit works perfectly for the Synaptics USB trackpad found in the detachable keyboard dock of my **Lenovo ThinkPad X1 Tablet Gen 3**, as well as for the USB trackpad found in the detachable type cover of my **Microsoft Surface Go 2** (the highest end model with the m3-8100Y processor). I'm still on Catalina 10.5.7 on both tablets. Without your commit, both trackpads won't work at all with VoodooI2C & the VoodooI2CHID satellite and work as a simple mouse if I remove the VoodooI2C kexts entirely (or use Karabiner Elements to handle them). VoodooI2C actually detects the precision trackpads and seems to activate them, but they won't work at all.
> As I don't have any device with an I2C trackpad on hand, I'm afraid I won't be able to help you check if the commit won't break other I2C trackpads, though.
> I confirm as well that the I2C touchscreens handled by VoodooI2CHID keep working on both tablets with your commit. The touchscreen is a Synaptics on the Lenovo and an ELAN on the Microsoft tablet.
> I have working battery status on both tablets and therefore the trackpad settings appear in system preferences. They appear to be shared with the touchscreen, though. That's the same issue as when you use a Magic Trackpad or Magic Mouse on a MacBook. The internal trackpad and the Magic Trackpad share the same settings.
> Please tell me if you need more information, logs or anything to help you getting this commit merged.

Hi, @jlempen, sorry to pin to here. Because I can't find another way to contact you.
I and another friend encountered some difficulties in solving the sleep/wake issue of surface go2 Hackintosh. 
https://www.reddit.com/r/hackintosh/comments/mwrzxg/intel_igpu_hdmiinternal_screen_problem/
https://github.com/kingo132/surface-go2-hackintosh/issues/1
I think it's a little hard for us to solve it. So I'm asking if you could look into this problem and give some clues about this.
Many thanks, appreciate it if you could help.

kprinssu:
@kingo132 Please keep discussions to VoodooI2C and the PR at hand. This is to keep the discussions easier to read through and less cluttered. Please contact the original some other way regarding general Hackintosh questions.

Kethen:
@lulujyc 
> Hello @Kethen ,
> I've tested this driver on the HP Elite X2 G4 trackpad, which is provided by ELAN instead of Synaptics (as for the case of X1 Tablet Evo). When the Multitouch HID Event Driver is being used, the trackpad does not work at all. I wonder if you wouldn't mind helping ELAN trackpad users. Another case showing the same symptom can be found here: https://github.com/acai66/lenovo-miix-520-hackintosh

sorry I don't have such hardware and currently I don't even have access to the keyboard dock tablet so I can't bang my head against those :(

lulujyc:
> @lulujyc
> 
> > Hello @Kethen ,
> > I've tested this driver on the HP Elite X2 G4 trackpad, which is provided by ELAN instead of Synaptics (as for the case of X1 Tablet Evo). When the Multitouch HID Event Driver is being used, the trackpad does not work at all. I wonder if you wouldn't mind helping ELAN trackpad users. Another case showing the same symptom can be found here: https://github.com/acai66/lenovo-miix-520-hackintosh
> 
> sorry I don't have such hardware and currently I don't even have access to the keyboard dock tablet so I can't bang my head against those :(

Hahaha ok, thank you for replying!

jlempen:
FYI, I just installed VoodooI2C and the VoodooI2CHID plugin compiled with this commit on a Dell Latitude 5290 2-in-1, which uses an Alps V6 Touchpad in its type cover connected via USB, and the trackpad works beautifully with all gestures now.
This commit has made all my tablet trackpads usable with full gesture support and the touchscreens keep working as before (Microsoft Surface Go 2 with an ELAN touchpad, Lenovo ThinkPad X1 Tablet Gen 3 with a Synaptics touchpad, and now Dell Latitude 5290 2-in-1 with an Alps touchpad).

ben9923:
@Kethen would you mind rebasing the branch against upstream?
That fork is unsynced with VoodooI2C, leading to scaling issues.

Anyway, we need more testers...

ben9923:
Just noticed that's the only place `VoodooI2CPrecisionTouchpadFeatureReport` is used, so it may be removed as well.

1Revenger1:
With this, the `VoodooI2CHIDDevice Multitouch HID Event Driver` personality is never used as well. Other personalities attach to the same usage pairs and have higher probe scores. Only thing it's used for right now is USB touchpads. Can be a different PR, or be left in for people who want to more easily test I guess.

ben9923:
@Kethen Would you mind rebasing this on top of master? It seems like there's a conflict there.
Also note the needed removal of the feature report class, it's never used with those changes.

I wanna get this merged in, and release a new version.

@1Revenger1 I think we can keep that personality there as a fallback, though _currently_ I see no reason for the derived personalities to fail :P

Kethen:
@ben9923 I rebased it but I don't have any access to my test machines/ioregistry any more so please test this before merging, especially with the new `<key>IOMatchCategory</key>`

kprinssu (approved):
This looks good to me.

PR #49, opened 2021-02-06
Fix unit exponent factoring for physical max values

ben9923:
Also scale values to 0.01 mm units.

### DIFF
--- a/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
@@ -23,12 +23,18 @@ OSDefineMetaClassAndStructors(VoodooI2CMultitouchHIDEventDriver, IOHIDEventServi
 
 AbsoluteTime last_multi_touch_event = 0;
 
-static int pow(int x, int y) {
-    int ret = 1;
-    while (y > 0) {
-        ret *= x;
-        y--;
+static int scientific_pow(UInt32 significand, UInt32 base, SInt32 exponent) {
+    UInt32 ret = significand;
+    while (exponent > 0) {
+        ret *= base;
+        exponent--;
     }
+
+    while (exponent < 0) {
+        ret /= base;
+        exponent++;
+    }
+
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -23,12 +23,18 @@ OSDefineMetaClassAndStructors(VoodooI2CMultitouchHIDEventDriver, IOHIDEventServi
 
 AbsoluteTime last_multi_touch_event = 0;
 
-static int pow(int x, int y) {
-    int ret = 1;
-    while (y > 0) {
-        ret *= x;
-        y--;
+static int scientific_pow(UInt32 significand, UInt32 base, SInt32 exponent) {
```
kprinssu, line 26: Can we not use any built-in to STL  for this function? 

I don't recall if cmath or the C++ equivalent has it.

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -468,6 +474,28 @@ void VoodooI2CMultitouchHIDEventDriver::handleStop(IOService* provider) {
...
+
+    UInt8 raw_unit_exponent = element->getUnitExponent();
+    if (raw_unit_exponent >> 3) {
+        raw_unit_exponent = raw_unit_exponent | 0xf0; // Raise the 4-bit int to an 8-bit int
+    }
+    SInt8 unit_exponent = *reinterpret_cast<SInt8*>(&raw_unit_exponent);
+
+    // Scale to 0.01 mm units
+    UInt32 unit = element->getUnit();
+    if (unit == 0x11) {        // Centimeters
```
kprinssu, line 488: Can we put 0x11 and 0x13 as header definitions as these are constants?

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
ben9923: I couldn't find one that matches. We could use floats, but I'm not sure if using floating point arithmetics is safe in the kernel. I know it isn't in Linux.

IIRC the former method was also our own to avoid linking issues between OS versions. Not sure using MacKernelSDK will currently solve those.

kprinssu: Ah okay. We lose precision and performance by using floats, let's not.

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
ben9923: Done.

kprinssu:
Thanks @ben9923

PR #51, opened 2021-02-20
Stop interrupts after power off, start before power on/reset

ben9923:
Devices with the I2C_HID_QUIRK_RESET_ON_RESUME quirk will wait for an interrupt after the device reset upon wake.
Interrupts were only enabled after the reset, making devices timeout waiting for them as a signal of a successful reset.

Devices with the I2C_HID_QUIRK_NO_IRQ_AFTER_RESET quirk were the primary reason for the previous change, making the bug hidden until it was recently reported.

Same behavior may be seen in the [Linux i2c-hid driver](https://github.com/torvalds/linux/blob/master/drivers/hid/i2c-hid/i2c-hid-core.c) (Well, it wouldn't work otherwise).

### DIFF
--- a/VoodooI2CHID/VoodooI2CHIDDevice.cpp
+++ b/VoodooI2CHID/VoodooI2CHIDDevice.cpp
@@ -414,9 +414,9 @@ IOReturn VoodooI2CHIDDevice::setPowerState(unsigned long whichState, IOService* 
         return kIOReturnInvalid;
     if (whichState == 0) {
         if (awake) {
-            stopInterrupt();
-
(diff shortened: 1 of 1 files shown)


kprinssu:
Looks good!

kprinssu (approved):

ben9923:
> Looks good!
> 
> This fixes [VoodooI2C/VoodooI2C#436](https://github.com/VoodooI2C/VoodooI2C/issues/436) correct?

Yes, it does :)
Verified by @michyprima and partially by @usr-sse2

PR #52, opened 2021-03-13
Fixes for USB-connected keyboard and trackpad

Xiashangning:
The keyboard patch is from `https://github.com/shdkpr2008/VoodooI2CHID.git`'s VoodooI2CMultitouchHIDEventDriver.cpp/hpp. Probably this is the solution for all USB-connected detachable keyboard&trackpad like Surface series, 2 in 1 laptops etc.

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -141,6 +141,26 @@
 			<key>IOProviderClass</key>
 			<string>IOHIDInterface</string>
 		</dict>
+		<key>VoodooI2CHIDDevice Keyboard HID Event Driver</key>
+		<dict>
+			<key>CFBundleIdentifier</key>
+			<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+			<key>DeviceUsagePairs</key>
+			<array>
+				<dict>
+					<key>DeviceUsage</key>
+					<integer>6</integer>
+					<key>DeviceUsagePage</key>
+					<integer>1</integer>
+				</dict>
+			</array>
+			<key>IOProbeScore</key>
+			<integer>200</integer>
+			<key>IOClass</key>
+			<string>VoodooI2CKeyboardHIDEventDriver</string>
+			<key>IOProviderClass</key>
+			<string>IOHIDInterface</string>
+		</dict>
 		<key>VoodooI2CHIDDevice Generic Mouse HID Event Driver</key>
 		<dict>
 			<key>CFBundleIdentifier</key>
--- a/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
@@ -16,12 +16,17 @@
 #include <IOKit/hid/AppleHIDUsageTables.h>
 
 #define SET_NUMBER(key, num) do { \
-tmpNumber = OSNumber::withNumber(num, 32); \
-if (tmpNumber) { \
-kbEnableEventProps->setObject(key, tmpNumber); \
-tmpNumber->release(); \
-} \
+    tmpNumber = OSNumber::withNumber(num, 32); \
+    if (tmpNumber) { \
+        kbEnableEventProps->setObject(key, tmpNumber); \
+        tmpNumber->release(); \
+    } \
 }while (0);
+
+// constants for processing the special key input event
+#define kHIDIncrVolume  0x01
+#define kHIDDecrVolume  0x02
+#define kHIDMute        0x04
 
 #define GetReportType(type)                                             \
 ((type <= kIOHIDElementTypeInput_ScanCodes) ? kIOHIDReportTypeInput :   \
@@ -140,7 +145,7 @@ void VoodooI2CMultitouchHIDEventDriver::handleInterruptReport(AbsoluteTime times
         digitiser.current_report++;
     }
     
-    handleKeboardReport(timestamp, report_id);
+    handleKeyboardReport(timestamp, report_id);
 }
 
 void VoodooI2CMultitouchHIDEventDriver::handleDigitizerReport(AbsoluteTime timestamp, UInt32 report_id) {
@@ -390,7 +395,7 @@ void VoodooI2CMultitouchHIDEventDriver::handleDigitizerTransducerReport(VoodooI2
         return;
 }
 
-void VoodooI2CMultitouchHIDEventDriver::handleKeboardReport(AbsoluteTime timeStamp, UInt32 reportID) {
+void VoodooI2CMultitouchHIDEventDriver::handleKeyboardReport(AbsoluteTime timeStamp, UInt32 reportID) {
     UInt32      volumeHandled   = 0;
     UInt32      volumeState     = 0;
     UInt32      index, count;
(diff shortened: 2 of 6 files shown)


kprinssu:
Thanks @Xiashangning, I'll be creating a PR review.

@ben9923 can you please also take a look.

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -596,6 +700,132 @@ IOReturn VoodooI2CMultitouchHIDEventDriver::parseDigitizerElement(IOHIDElement*
...
+                     
+                     SET_NUMBER(kIOHIDKeyboardEnabledEventEventTypeKey, kIOHIDEventTypeKeyboard);
+                     SET_NUMBER(kIOHIDKeyboardEnabledEventUsagePageKey, kHIDPage_KeyboardOrKeypad);
+                     SET_NUMBER(kIOHIDKeyboardEnabledEventUsageKey, kHIDUsage_KeyboardPower);
+                     
+                     setProperty(kIOHIDKeyboardEnabledEventKey, kbEnableEventProps);
+                     setProperty(kIOHIDKeyboardEnabledByEventKey, kOSBooleanTrue);
+                     setProperty(kIOHIDKeyboardEnabledKey, value ? kOSBooleanTrue : kOSBooleanFalse);*/
+                    
+                    kbEnableEventProps->release();
```
kprinssu, line 773: `OSSafeReleaseNULL` would be better here.

kprinssu: In fact why even allocate `kbEnableEventProps` if it's not being used?

kprinssu (changes requested):

Xiashangning:
Got it, in fact it is my bad not to examine through the code before I submit this PR, I just `copy` what in his code without thinking carefully what he has done, since I have been using this patch for quite a few days and no KP is met, I will try to ameliorate it and then test it to see if something unexcepted would happen.

kprinssu:
@Xiashangning No problem, it's mostly stylistic an d nothing major in general, but a huge thanks for creating the PR!

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -378,6 +390,97 @@ void VoodooI2CMultitouchHIDEventDriver::handleDigitizerTransducerReport(VoodooI2
...
+         }
+         }*/
+        
+        dispatchKeyboardEvent(timeStamp, usagePage, usage, value);
+    }
+    
+    // RY: Handle the case where Vol Increment, Decrement, and Mute are all down
+    // If such an event occurs, it is likely that the device is defective,
+    // and should be ignored.
+    if ( (volumeState != 0x7) && (volumeHandled != 0x7) ) {
```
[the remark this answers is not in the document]
Xiashangning: Like a single .h header file? I would name it `VoodooI2CHIDKeyboardConstants.h`, but there are only three constants right now... So maybe `#define kHIDIncVolume 0x01`?

ben9923:
Thanks for the PR :)

Until I can properly review it, I must say putting the keyboard code in the multitouch event driver seems like the wrong idea.
Ideally another event driver should attach to the same IOHIDInterface and handle keyboard reports.

BTW, can you attach a copy of IOReg from your machine? I'd like to see how the matching can be done there, when separated into a dedicated keyboard event driver.

Xiashangning:
@kprinssu I have made the required improvement, the magic constants are define at the head of 
`VoodooI2CMultitouchHIDEventDriver.cpp`.
@ben9923 Here it is. It's under XHC/HS08
[Surface Pro 7 ioreg.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/6157439/Surface.Pro.7.ioreg.zip)

BTW, you can see that there is another I2C controller presented in my IO reg(`I2C4`), it has a device id of `34c5` which is not supported by VoodooI2C. However, it has an ambient light sensor attached under it and the code from Linux is very simple to port. I have searched through VoodooI2C but cannot find how the `VoodooI2CPCILakeController` is matching against these devices, I can only find 0x34e8 in the IOKitPersonalities. I would like to try adding 34c5 and test if it is compatible, how should I do? Thanks

ben9923:
@Xiashangning Thanks for the IOReg :)
So indeed that driver can match usage page 1, usage 6 (Which means keyboard).
I'm not sure if another driver can just match under the same IOHIDInterface, or we should make some changes to allow both to load at the same time...
If you remove the `VoodooI2CMultitouchHIDEventDriver` personality from VoodooI2CHID's Info.plist, does any other driver load instead? Does the keyboard work in that case? Please attach an IOReg with it disabled :)

Anyway, the keyboard code cannot be a part of the multitouch code. They are completely unrelated to each other.

As for `8086:34c5`, you can just add `34cX` to the `VoodooI2CPCILakeController` personality in the Info.plist.
For example, you can see `0x34e88086&0xFFFCFFFF` there. It'll match any device with PCI ID `8086:34eX`, Just add `0x34c58086&0xFFFCFFFF` to the same PCI matching.

Xiashangning:
@ben9923 Hello, I have made a dedicated keyboard driver `VoodooI2CKeyboardHIDEventDriver`, using your recommended match pairs. As you can see in the attached IOReg, only the Keyboard Driver is activated, macOS fails to load multitouch driver...
I have committed my changes. Frankly speaking, I do not know too much about kext development, still learning :) So please take a look at my code and tell me if I have done anything wrong.
[Surface Pro 7 dedicated Keyboard Driver.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/6177755/Surface.Pro.7.dedicated.Keyboard.Driver.zip)

ben9923:
@Xiashangning Cool :)
Did you also check the case when VoodooI2C doesn't load under it at all? Does any Apple keyboard driver attach to the device?

Seems like only one event driver may load under an `IOHIDInterface`. Someone should find the appropriate method to make use of both, either by a third event driver that decorates the other two, or by any other way that would allow both to attach at the same time.

I just checked how my keyboard + mouse receiver works - It simply exposes multiple `IOHIDInterface`s...

VoodooI2CHID/Info.plist
```diff
@@ -141,6 +141,26 @@
...
+				<dict>
+					<key>DeviceUsage</key>
+					<integer>6</integer>
+					<key>DeviceUsagePage</key>
+					<integer>1</integer>
+				</dict>
+			</array>
+			<key>IOProbeScore</key>
+			<integer>200</integer>
+			<key>IOClass</key>
```
1Revenger1, line 159:
```suggestion
                        <key>IOMatchCategory</key>
                        <string>VoodooI2CKeyboardHIDEventDriver</string>
			<key>IOClass</key>
```
The string can be whatever you want, but having different categories for matching should allow multiple services to probe + attach

kingo132:
This patch does not support caps lock led on the surface type cover. I read the code of VoodooPS2 tries to find how to toggle this caps led. But it seems very different from ApplePS2KeyboardDevice (used by VoodooPS2) to IOHIDEventService (used by your patch). I tried to add some code to toggle caps led. They all failed. Could you please add this led toggle?

therealjbenam:
Any news on getting this merged in VoodooI2C, @Xiashangning ?

I have compiled the kext with your patches and it works wonderfully on my Surface Pro 6. Finally no more KPs when closing the lid (which means I can now also use it in "docked" mode, with the lid closed and an external monitor and keyboard, yay)!

Xiashangning:
@therealjbenam sorry, it seems that there is concern about putting keyboard related codes in trackpad&i2c kext(which is the nature of this repo),so instead of merging it, i have already created my repo forked from this repo and add them myself. Since my kext is dedicated for SP users, mixing these codes together is fine for me. :D
You can use that kext instead, i will update it along with voodooi2c. 
Hence this pr will be closed(sorry for leaving this pr for so long, i forgot to close it…)

PR #55, opened 2021-09-07
Prevent VoodooI2CHID from attaching to PS2 devices

kprinssu:
Fixes https://github.com/VoodooI2C/VoodooI2C/issues/469


ben9923:
Wouldn't such change break USB trackpads/touchscreens?

kprinssu:
@ben9923 Yes, it should go into `VoodooI2CKeyboardHIDEventDriver`.....

PR #57, opened 2021-10-08
Fix some issues regarding getReport and setReport

juico:
When i was trying to fix the Alps4TUSB from blankmac i ran into some issues with setReport and getReport. 

For the setReport i looked up the linux i2c-hid code and there they skipped the first byte of the featureReport as it contains the report_id. Later in the code they add the report_id to the message again. In the voodooI2CHID it is not removed in the first place but is added again like in the linux driver.  This introduces a double report_id and thus corrupts the featureReport.

For the getReport function there seems to be a issue that the last two bytes of the report are not written. Fixed by making the readbuffer two bytes bigger and writing the complete result back into the report memoryDescriptor.

These two fixes make the communication part of blankmac's code work. As his code works when using USB and not using I2C i am convinced these fixes are needed to make voodooI2CHID work better.

### DIFF
--- a/VoodooI2CHID/VoodooI2CHIDDevice.cpp
+++ b/VoodooI2CHID/VoodooI2CHIDDevice.cpp
@@ -347,12 +347,17 @@ IOReturn VoodooI2CHIDDevice::setReport(IOMemoryDescriptor* report, IOHIDReportTy
     UInt16 size;
     UInt16 arguments_length;
     UInt8 report_id = options & 0xFF;
-    UInt8* buffer = (UInt8*)IOMalloc(report->getLength());
-    report->readBytes(0, buffer, report->getLength());
-
+    UInt16 report_length = report->getLength();
+    UInt8* buffer = (UInt8*)IOMalloc(report_length);
+    
+    report->readBytes(0, buffer, report_length);
+    if(report_id == buffer[0]){
+        buffer++;
+        report_length--;
+    }
     size = 2 +
     (report_id ? 1 : 0)     /* reportID */ +
-    report->getLength()     /* buf */;
+    report_length     /* buf */;
 
     arguments_length = (report_id >= 0x0F ? 1 : 0)  /* optional third byte */ +
     2                                               /* dataRegister */ +
@@ -375,7 +380,7 @@ IOReturn VoodooI2CHIDDevice::setReport(IOMemoryDescriptor* report, IOHIDReportTy
     if (report_id)
         arguments[idx++] = report_id;
 
-    memcpy(&arguments[idx], buffer, report->getLength());
+    memcpy(&arguments[idx], buffer, report_length);
 
     UInt8 length = 4;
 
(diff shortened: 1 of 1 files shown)


mishurov:
In Linux, my primary OS, in userspace, I can do this:
```
#define INPUT_MODE_REPORT_ID 0x3
#define INPUT_MODE_TOUCHPAD 0x03
#define LATENCY_MODE_REPORT_ID 0x7
#define LATENCY_MODE_NORMAL 0x00

unsigned char buf[2];
buf[0] = INPUT_MODE_REPORT_ID;
buf[1] = INPUT_MODE_TOUCHPAD;
res = ioctl(fd, HIDIOCSFEATURE(3), buf);
```
Just using the file desctiptor `/dev/hidraw0` and writing into it. In kernel space it is a bit more complicated since it uses `struct hid_device` and its functionality. I am not quite sure what goes wrong in your case. But I'll try your fix when i have some more time.

ben9923:
@juico Just found the time to properly review it.

Good catch on the `getReport` fix, looking at the I2C-HID spec clearly shows the problem (two _extra_ bytes for `length` at the beginning).

Regarding `setReport`, I see that the Linux kernel just checks if the first byte is nonzero, then 'marks' it as `reportID` and removes it from the buffer. Would you explain why your code behaves differently? :)

Also, any idea why it's even transferred as part of the report in the first place? I see macOS is using the `options` argument.
In Linux, when the `reportID` argument is 'introduced' in the code flow, it's eliminated from the buffer.

Thanks again! I would be happy to merge this in soon, just making sure we don't break anything by accident.

juico:
@ben9923 Thanks for checking it out. For the setReport i remember comparing the first byte against the supplied reportID from the options argument as this is expected to be the case. Just checking for nonzero byte would also work, perhaps that would make more sense and just return a error if the reportID in the first byte doesnt match that of the one supplied in the options argument. Some of the other differences are due to the differences in the types of buffer used but it is possible there is a cleaner solution.

Not sure why it is a part of the report itself. I tried to figure that out but had some trouble finding clear documentation about it. So i ended up using the webHID interface of chrome to call setReport and checked how it looked like by logging the reports fed to setReport. That lead to the conclusion that internaly the reportID is placed a the beginning of the report itself and in the options argument.

In terms of breaking stuff it seemed that by merging the master branch to my fork it broke some things for me, havent had time to check out why this is the case. Perhaps if i have some spare time in the weekend so i can figure out why it breaks.

VoodooI2CHID/VoodooI2CHIDDevice.cpp
```diff
@@ -347,12 +347,17 @@ IOReturn VoodooI2CHIDDevice::setReport(IOMemoryDescriptor* report, IOHIDReportTy
     UInt16 size;
     UInt16 arguments_length;
     UInt8 report_id = options & 0xFF;
-    UInt8* buffer = (UInt8*)IOMalloc(report->getLength());
-    report->readBytes(0, buffer, report->getLength());
-
+    UInt16 report_length = report->getLength();
```
kprinssu, line 350: The report_length may not always be 16 bytes in length.

See: https://developer.apple.com/documentation/kernel/iomemorydescriptor/1442036-getlength


We may have overflow issues here,

kprinssu (approved):

VoodooI2CHID/VoodooI2CHIDDevice.cpp
kprinssu: We'll tackle this in a follow up PR.

PR #58, opened 2021-12-23
Fix report parsing for devices with multi-mode (trackpad or touchscreen) capabilities

gvkt:
This provides a fix for correctly parsing the report descriptor for the desired mode in a multi-mode device such as the Asus Screenpad (e.g., as trackpad or as touch screen).

In such devices the same elements (e.g., contact count) can exist in multiple sections in the report descriptor. While parsing the report descriptor, the current code assumes a single mode and parses all elements that can work in any mode. Depending on the order in which the record descriptor elements are parsed, the wrong object may be cached for reporting for the mode in which the satellite driver is using the device in. 

This results in symptoms such as multi-touch not being available for such a device as the contact count element object cached may be for the wrong mode for the mode in which the device is being used. That cached object will then not update as reports come in for the device in desired mode.

The fix allows the specialized device event drivers such as PrecisionTrackpad or TouchScreen to control the parsing and caching for their respective modes only.

This fix has no known side-effects (not tested with pen/stylus capability) and has fixed multi-touch unavailability problem for the Asus Screenpad working as a trackpad and the second screen of Asus Duo laptops working as touch screen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,13 @@
+# VoodooI2CHID
+
+This is a fork with a fix for the Precision Trackpad and Touch Screen Trackpad satellites to work correctly for devices that can work with multiple modes (trackpad or touch screen).
+
+An example of such a device would be the Screenpads in many ASUS laptops. 
+
+Without this fix, such devices only work in single finger mode.
+
+The root cause is in the parsing of the report descriptors to maintain a list of supproted elements to use when an input report is received. The original code assumed a single mode for any device. But this could result in incorrect caching of element objects for a record descriptor which contained common elements in multiple modes.
+
+The fix allows the Precision Trackpad and Touch Screen Event Drivers to narrow the parsing of record descriptors to their respective modes only.
+
+This has solved the multi-touch problem with Screenpad devices such as The Goodix GDX1515 amongst others.
--- a/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
@@ -431,7 +431,7 @@ bool VoodooI2CMultitouchHIDEventDriver::handleStart(IOService* provider) {
     if (!digitiser.transducers)
         return false;
 
-    if (parseElements() != kIOReturnSuccess) {
+    if (parseElements(0) != kIOReturnSuccess) {
         IOLog("%s::%s Could not parse multitouch elements\n", getName(), name);
         return false;
     }
@@ -597,7 +597,7 @@ IOReturn VoodooI2CMultitouchHIDEventDriver::parseDigitizerElement(IOHIDElement* 
     return kIOReturnSuccess;
 }
     
-IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements() {
+IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements(UInt32 forUsage) {
     int index, count;
 
     OSArray* supported_elements = OSDynamicCast(OSArray, hid_device->getProperty(kIOHIDElementKey));
@@ -615,16 +615,24 @@ IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements() {
         if (element->getUsage() == 0)
             continue;
 
-        if (element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_Pen)
-            || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen)
-            || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchPad)
-            || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_DeviceConfiguration)
-            )
-            parseDigitizerElement(element);
-        
-        if (multitouch_interface && element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen))
-            multitouch_interface->setProperty(kIOHIDDisplayIntegratedKey, kOSBooleanTrue);
+        if (forUsage) {
+            if (element->conformsTo(kHIDPage_Digitizer, forUsage)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_DeviceConfiguration)
+                )
+                parseDigitizerElement(element);
+        } else {
+            if (element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_Pen)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchPad)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_DeviceConfiguration)
+                )
+                parseDigitizerElement(element);
+            }
+        if (!forUsage || (forUsage == kHIDUsage_Dig_TouchScreen)) {
+            if (multitouch_interface && element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen))
+                multitouch_interface->setProperty(kIOHIDDisplayIntegratedKey, kOSBooleanTrue);
         }
+    }
     
 
     if (digitiser.styluses->getCount() == 0 && digitiser.fingers->getCount() == 0)
--- a/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.hpp
+++ b/VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.hpp
@@ -178,13 +178,7 @@ class EXPORT VoodooI2CMultitouchHIDEventDriver : public IOHIDEventService {
 
     IOReturn parseDigitizerTransducerElement(IOHIDElement* element, IOHIDElement* parent);
 
-    /* Parses all matched elements
-     *
-     * @return *kIOReturnSuccess* on successful parse, *kIOReturnNotFound* if the matched elements are not supported, *kIOReturnError* otherwise
-     */
-
-    IOReturn parseElements();
-
+    
     /* Postprocessing of digitizer elements
      *
      * This function is mostly copied from Apple's own HID Event Driver code. It is responsible for cleaning up malformed report descriptors as well as setting some miscellaneous properties.
@@ -266,6 +260,13 @@ class EXPORT VoodooI2CMultitouchHIDEventDriver : public IOHIDEventService {
     bool should_have_interface = true;
 
     virtual void forwardReport(VoodooI2CMultitouchEvent event, AbsoluteTime timestamp);
+    
+    /* Parses all matched elements
+     *
+     * @return *kIOReturnSuccess* on successful parse, *kIOReturnNotFound* if the matched elements are not supported, *kIOReturnError* otherwise
+     */
+
+    virtual IOReturn parseElements(UInt32 forUsage=0);
 
  private:
     SInt32 absolute_axis_removal_percentage = 15;
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
@@ -49,6 +49,7 @@ class EXPORT VoodooI2CPrecisionTouchpadHIDEventDriver : public VoodooI2CMultitou
     IOReturn setPowerState(unsigned long whichState, IOService* whatDevice) override;
 
  protected:
+    IOReturn parseElements(UInt32 forUsage=0) override;
  private:
     bool ready = false;
 
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
@@ -49,6 +49,10 @@ bool VoodooI2CPrecisionTouchpadHIDEventDriver::handleStart(IOService* provider) 
     return true;
 }
 
+IOReturn VoodooI2CPrecisionTouchpadHIDEventDriver::parseElements(UInt32) {
+    return super::parseElements(kHIDUsage_Dig_TouchPad);
+}
+
 IOReturn VoodooI2CPrecisionTouchpadHIDEventDriver::setPowerState(unsigned long whichState, IOService* whatDevice) {
     if (whatDevice != this)
         return kIOReturnInvalid;
--- a/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
@@ -111,6 +111,10 @@ void VoodooI2CTouchscreenHIDEventDriver::checkRotation(IOFixed* x, IOFixed* y) {
     }
 }
 
+IOReturn VoodooI2CTouchscreenHIDEventDriver::parseElements(UInt32) {
+    return super::parseElements(kHIDUsage_Dig_TouchScreen);
+}
+
 bool VoodooI2CTouchscreenHIDEventDriver::checkStylus(AbsoluteTime timestamp, VoodooI2CMultitouchEvent event) {
     //  Check the current transducers for stylus operation, dispatch the pointer events and return true.
     //  At this time, Apple has removed all methods of handling additional information from the event driver.  Only x, y, buttonstate, and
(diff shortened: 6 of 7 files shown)


VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -597,7 +597,7 @@ IOReturn VoodooI2CMultitouchHIDEventDriver::parseDigitizerElement(IOHIDElement*
     return kIOReturnSuccess;
 }
     
-IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements() {
+IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements(UInt32 forUsage) {
```
kprinssu, line 600: Can you please consider a name change to just usage (see https://developer.apple.com/documentation/kernel/iohidelement/1426848-conformsto)

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -431,7 +431,7 @@ bool VoodooI2CMultitouchHIDEventDriver::handleStart(IOService* provider) {
     if (!digitiser.transducers)
         return false;
 
-    if (parseElements() != kIOReturnSuccess) {
+    if (parseElements(0) != kIOReturnSuccess) {
```
kprinssu, line 434: What does `0` imply here, it's more or less a magic constant? Can you use a constant from IO Kit?

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.hpp
```diff
@@ -266,6 +260,13 @@ class EXPORT VoodooI2CMultitouchHIDEventDriver : public IOHIDEventService {
     bool should_have_interface = true;
 
     virtual void forwardReport(VoodooI2CMultitouchEvent event, AbsoluteTime timestamp);
+    
+    /* Parses all matched elements
+     *
+     * @return *kIOReturnSuccess* on successful parse, *kIOReturnNotFound* if the matched elements are not supported, *kIOReturnError* otherwise
+     */
+
+    virtual IOReturn parseElements(UInt32 forUsage=0);
```
kprinssu, line 269: Please document what `forUsage` parameter is and does.

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
```diff
@@ -49,6 +49,7 @@ class EXPORT VoodooI2CPrecisionTouchpadHIDEventDriver : public VoodooI2CMultitou
     IOReturn setPowerState(unsigned long whichState, IOService* whatDevice) override;
 
  protected:
+    IOReturn parseElements(UInt32 forUsage=0) override;
```
kprinssu, line 52: Please replace the magic constant `0` with a header declaration or a proper equivalent from IO Kit.

VoodooI2CHID/VoodooI2CMultitouchHIDEventDriver.cpp
```diff
@@ -615,16 +615,24 @@ IOReturn VoodooI2CMultitouchHIDEventDriver::parseElements() {
...
+        } else {
+            if (element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_Pen)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchPad)
+                || element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_DeviceConfiguration)
+                )
+                parseDigitizerElement(element);
+            }
+        if (!forUsage || (forUsage == kHIDUsage_Dig_TouchScreen)) {
+            if (multitouch_interface && element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen))
```
kprinssu, line 632: Why not roll the if statement into one as:
`( *!forUsage || (forUsage == kHIDUsage_Dig_TouchScreen)) && multitouch_interface && element->conformsTo(kHIDPage_Digitizer, kHIDUsage_Dig_TouchScreen))`

Perhaps adding a multi-line condition would be more readable but I am not a fan of nested if statements.

kprinssu (changes requested):

README.md
```diff
@@ -0,0 +1,13 @@
+# VoodooI2CHID
+
+This is a fork with a fix for the Precision Trackpad and Touch Screen Trackpad satellites to work correctly for devices that can work with multiple modes (trackpad or touch screen).
```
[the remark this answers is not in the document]
gvkt: I wrote the README to explain what the fork was about and it automatically got included in the changes. Still haven't figured out how to withdraw or reject an item for this purpose. New to this.

gvkt:
I am fine with whatever stylistic changes you guys want. That is neither my strong point nor productive use of my time. Spent far too much time already with the steep learning curve starting from zero knowledge to figure out the problem and a proof of concept solution to make sure it works. Didn't plan on becoming a Voodoo developer.

I will leave it to whoever is motivated and have the time on how best to incorporate into the existing code base if they believe this problem is worth fixing. Have a bunch of issues still to fix with my Hackintosh before I actually use it for productive work. :) 

Hope the time I spent was worth it to more than just me.

kprinssu:
No worries @gvkt 

Thanks for the PR. I'll reopen it but myself or @ben9923 can change it. I'll reopen this so we can actually track the changes.

ben9923:
@gvkt Thanks for your work!

We might be able to make it a bit cleaner, by checking whatever usage was the driver matched by, but I'm not sure how it'll work for the generic multitouch driver.
I'll try to work on that :)

gvkt:
@ben9923  I thought a little bit about the solution before the quick fix for my specific use case. So, I will put them down here to guide any solution.

I kept the changed code in MultiTouchEventDriver separated visually so that its behavior and usage in existing hardware where it is attached directly rather than the PTP or TouchScreen specializations is exactly as before and minimal perturbation without creating any regression issues. I see this as just a quick short term hack.

To be conceptually clean and correct, MultiTouchEventDriver code requires a non-trivial rewrite to make the "digitizer" variable as an indexed array for each "collection" in the report descriptor, indexed by the unique Report ID. So there would be separated digitizer object lists for mouse collection, trackpad collection, touchscreen collection, etc as may be present in the type descriptor each indexed with its own unique Report ID. During parsing, the child element objects would go into the corresponding digitizer collection rather than be squished into a single list as it happens now with element objects from different collections overwriting each other depending on the order in which they are processed.

When a report comes in for any collection, the report id would be used to pick the cached digitizer object collection that would reflect the data from that report.

If this is done, then no change would be necessary for PTP or Touchscreen event drivers. The changes in this PR to the specializations would be unnecessary except perhaps as an optimization to narrow the parsing/caching to only a single collection when the usage mode is known and static.

So, I see this quick fix suitably cleaned up as a short term temporary solution to fix existing issues with contemporary laptops. I don't see any future laptop use cases likely that there would be a multi-touch device that is multi-modal but neither a precision trackpad nor a touch screen. So, in theory, the generic MultiTouch event driver wouldn't attach directly to any device except perhaps for some legacy devices that aren't PTP compliant.

The longer term solution as described above should be addressed in a rethink of which and how event drivers will be assigned to a device in the first place. In laptops like the Zenbook and likely in most future laptops following Windows guidelines, the trackpad, the touch screen, etc will all be I2C HID devices differentiated in their properties only with a product ID (and in some cases vendor ID). So, the current scheme of automatic selection based on some probe score will not work correctly. The OOB kext even with this quick fix would not be usable for new Zenbook laptops such as mine without manually modifying the Info.plist to force the correct event driver to each device based on matching the product ID or some other differentiating property unrelated to their desired function.

Moreover, as currently happens in Zenbooks under Windows, that mode selection for a device is actually dynamic and user selectable during a booted session (e.g., screenpads). So a longer term solution such as the one described above would even make it potentially possible for a dynamic selection in a Hackintosh if use cases presented themselves in this context.

Hope that helps. Happy to test any beta releases in the future with my laptops for the fixes related to this.

narcyzzo:
can this be merged sometimes? many guys having touchscreens have issues with multi-mode so that will be great.

kprinssu:
> can this be merged sometimes? many guys having touchscreens have issues with multi-mode so that will be great.

Not as it is as there are a lot of glaring issues. @gvkt has already mentioned they will not be working on this further. I personally do not have the time to make the changes neither, but follow up PRs are welcome.

narcyzzo:
to be dead honest this is working. I tested it, so it just need code style poloshing isnt?

kprinssu:
> to be dead honest this is working. I tested it, so it just need code style poloshing isnt?

Cool. Yes, give the PR comments a go and if you want to work on the changes, great. As I said before, PRs are always welcome.

1Revenger1:
So I was looking to possibly pick this up. I'm not sure that switching dynamically between a trackpad and touchscreen is possible, especially with the current VoodooI2CHID architecture.  VoodooI2C in the Info.plist looks at usagepage/usage and only will take the one which has the highest score. You might be able to get both to attach if they filtered their own report IDs, as well as giving each their own IOMatchCategory. There would still be the issue of configuring which one was enabled/disabled though, which would require probably some sort of userspace application and Asus specific kext which looked at hotkeys which would both be outside the scope of this project.

Is there a way to configure the screenpad to act as a trackpad or touchscreen, or does it send both types of reports at the same time? I notice that they have different report IDs.

I think for now, the option of just filtering for the Touchpad or Touchscreen elements works just fine. If the Precision Trackpad event driver has the highest probe score, that should be the one used by default, and should be fine for hopefully 80% of cases? I'm not sure how this would affect stylus though.

I'm willing to pick this up (I'm bored), though I don't have anyway to test so I'd appreciate a couple testers once I post a build.

As an aside, does anyone know if devices like the Asus Duo work fine? I think they have two touchscreens plus whatever the touchpad does.

gvkt:
I have probably looked at the ASUS specifics of this more than anyone else at the moment. While I am not a hard coder, happy to share my findings, architecture suggestions and beta test.

> So I was looking to possibly pick this up. I'm not sure that switching dynamically between a trackpad and touchscreen is possible, especially with the current VoodooI2CHID architecture. 

The only way this might work as I have indicated in another thread is having a single driver that attaches to all and every i2c HID device and then acts as a dynamic router to "plugin" drivers based on report id. This itself wouldn't care about what the device is configured as or how one might switch between them which would be application level switching that someone else can provide. But this is a significant rewriting of the architecture. This would neatly solve the issues below and would be backward compatible.

> VoodooI2C in the Info.plist looks at usagepage/usage and only will take the one which has the highest score. You might be able to get both to attach if they filtered their own report IDs, as well as giving each their own IOMatchCategory. There would still be the issue of configuring which one was enabled/disabled though, which would require probably some sort of userspace application and Asus specific kext that looked at hotkeys. These would both be outside the scope of this project.
> 
From what I have seen here, increasingly all of these I2C HID input devices are going to look the same at the driver level and so the current match scoring is not going to work. These devices whether the Screenpad 2.0 that is primarily a touchpad or the secondary screen on duos which is primarly a touch screen will match the same driver (whatever it is). So in my case, the multi-touch screen driver will attach to both Screenpad 2.0 and to the touch (main) screen. If you fiddle around with weights to make PTP driver attach to the Screenpad, then something else will attach to the main screen stopping it from working as a touch screen.

My suggestion for doing this in simple steps:

1. Merge the code I have with whatever stylistic changes into the main. This will work as is for the typical trackpad touch-screen configurations.
2. For Zenbooks with Screenpad and Duos with secondary screen, a manual assignment is needed in addition (and independent of the above) to force the assignment of the correct driver for each. I have done this by modifying the info.plist to match product id. But a solution where this can be done with perhaps a boot argument is probably a more usable solution.

The above will take care of all of the devices for now with minimal changes.

The flow re-architecting could be a longer term goal.

1Revenger1:
> The only way this might work as I have indicated in another thread is having a single driver that attaches to all and every i2c HID device and then acts as a dynamic router to "plugin" drivers based on report id. This itself wouldn't care about what the device is configured as or how one might switch between them which would be application level switching that someone else can provide. But this is a significant rewriting of the architecture. This would neatly solve the issues below and would be backward compatible.

I agree to an extent. It looks like Linux has a quirk which allows multiple drivers per HID device to appear. I'm fairly certain that we would only need to keep track of plugins per HID device, correct? Also does the device report both Trackpad and Touchscreen reports at the same time, or is that configured when setting the mode of the device (ie 0x3 for touchpad and 0x2 for touchscreen).

> From what I have seen here, increasingly all of these I2C HID input devices are going to look the same at the driver level and so the current match scoring is not going to work. These devices whether the Screenpad 2.0 that is primarily a touchpad or the secondary screen on duos which is primarly a touch screen will match the same driver (whatever it is). So in my case, the multi-touch screen driver will attach to both Screenpad 2.0 and to the touch (main) screen. If you fiddle around with weights to make PTP driver attach to the Screenpad, then something else will attach to the main screen stopping it from working as a touch screen.

I think the match score does sort of work for most devices. Most devices released today have a regular trackpad and a touchscreen. It's mostly Asus with these confusing configurations lol. Just for clarity, screenpads are represented as a single HID device which reports both the Touchscreen and Touchpad usages, correct? What does a Duo device look like to the OS? Also what about that crazy calculator thing?

Sorry for all the questions. Just trying to figure out if it's just screenpads that have the issue or other Asus models as well.

williambj1:
> What does a Duo device look like to the OS?

My friend owns a first gen Zephyrus Duo and I installed macOS on that device a year ago (unfortunately I don't have access to that device anymore). The secondary screen at the bottom works as a normal touchscreen and everything is working as expected (multi finger input, gestures, etc) with the latest release at that time.

> Also what about that crazy calculator thing?

I personally own an 8th gen Intel Zephyrus S and it has one of those numpad/trackpad on it. This driver[1] works in linux. Looking at the source code it seems that a i2c command is need to activate the lights for the numbers and the touch input needs to be divided into regions for each num key.

\[1\]: https://github.com/mohamed-badaoui/asus-touchpad-numpad-driver/blob/main/asus_touchpad.py#L153

narcyzzo:
from what i know Duo devices has different I2C touchscreen its, ELAN1200, see this here : 
https://github.com/deniro98/hackintosh-asus-zenbook-duo-ux482ea

narcyzzo:
Asus Duo devices probably have another I2C device? Like here, ELAN1200 
https://github.com/deniro98/hackintosh-asus-zenbook-duo-ux482ea
@deniro98 mayby u can told something more about how this screen is working?

1Revenger1:
Thanks. I think I have a somewhat better insight into the issue now. I think for now I'll work through the changes requested above, and come back at a later time. I'll see if I can think up of something better.

Also, I've been looking through IOHIDFamily and found the property `kIOHIDMultipleInterfaceEnabledKey`, which seems to allow an IOHIDInterface to appear for every top level collection in the report descriptor... Pretty sure 3/4 interfaces would appear for the Screenpad, which sounds like a lot of chaos (Mouse/Touchpad/Touchscreen/Some vendor specific device). If anyone knows more about this property, feel free to say something.

~~If someone feels like breaking things and is crazy, it'd be interesting to see what happens if the multiple interface key property is set in VoodooI2CHIDDevice::handleStart before super::handleStart is called.~~

gvkt:
> I agree to an extent. It looks like Linux has a quirk which allows multiple drivers per HID device to appear. I'm fairly certain that we would only need to keep track of plugins per HID device. Does the device report both Trackpad and Touchscreen reports at the same time, or is that configured when setting the mode of the device (ie 0x3 for touchpad and 0x2 for touchscreen)?

The report descriptor contains all the collections it will report on but I believe the reporting is only for the mode it is set to.

> Just for clarity, screenpads are represented as a single HID device which reports both the Touchscreen and Touchpad usages, correct? 

Yes

> What does a Duo device look like to the OS? Also what about that crazy calculator thing?

Both screenpads and duo screens look like multiple video displays and multiple I2C HID devices. Two separate I2C HID devices (or 3 if they also have a trackpad in addition to two screens like the Zenbook Pro Duo).

The calculator, etc are all software running on the second display as a touch screen nothing to do with the devices themselves. There is also a software trackpad overlay
to temporarily use it as a trackpad without switching modes. This is all at application level.

From driver perspective, there is a function key to switch the mode to pure trackpad (relative co-ordinates) or to display touch screen (absolute co-ordinates) or completely disable it. Linux cannot do this. Only on Windows.

kprinssu:
I am closing this PR in favour of https://github.com/VoodooI2C/VoodooI2CHID/pull/59

1Revenger1:
> Both screenpads and duo screens look like multiple video displays and multiple I2C HID devices. Two separate I2C HID devices (or 3 if they also have a trackpad in addition to two screens like the Zenbook Pro Duo).
> 
> From driver perspective, there is a function key to switch the mode to pure trackpad (relative co-ordinates) or to display touch screen (absolute co-ordinates) or completely disable it. Linux cannot do this. Only on Windows.

Cool, that makes sense. Instead of worry about multiple plugins possibly, could we have the multitouch plugin check for both types of collections and then just choose one? If we had one event driver that handled everything, then we wouldn't need to bother with trying to attach multiple plugins and switching between them.

This also means we could have a userspace program potentially set a property and have the trackpad switch modes real quick.

PR #60, opened 2022-02-22
Add report ID to set input mode command, revert input_mode->setValue

1Revenger1:
This should still work with USB devices since the format was fixed.

input_mode->SetValue seems to break completely with I2C on Catalina. I wasn't able to figure out why it was broken, but this was reported to work. User's input mode report had an id of 3 (same value as Touchpad mode). It would be good to see if this works with USB devices.
~~[VoodooI2CHID+reportID.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/8119470/VoodooI2CHID%2BreportID.zip)~~
[VoodooI2CHID+CondReportID.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/8128272/VoodooI2CHID%2BCondReportID.zip)

### DIFF
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
@@ -12,9 +12,15 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
-    // We should really do this using `input_mode_element->setValue(INPUT_MODE_TOUCHPAD)`.
-    // This appears to not work on Catalina or older though (Works in Big Sur/Monterey)
+    if (version_major > CATALINA_MAJOR_VERSION) {
+        // Update value from hardware so we can rewrite mode when waking from sleep
+        digitiser.input_mode->getValue(kIOHIDValueOptionsUpdateElementValues);
+        digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);
+        ready = true;
+        return;
+    }
     
+    // TODO: setValue appears to not work on Catalina or older
     VoodooI2CPrecisionTouchpadFeatureReport buffer;
     buffer.reportID = digitiser.input_mode->getReportID();
     buffer.value = INPUT_MODE_TOUCHPAD;
--- a/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
@@ -12,8 +12,19 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
-    digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);
+    // We should really do this using `input_mode_element->setValue(INPUT_MODE_TOUCHPAD)`.
+    // This appears to not work on Catalina or older though (Works in Big Sur/Monterey)
+    
+    VoodooI2CPrecisionTouchpadFeatureReport buffer;
+    buffer.reportID = digitiser.input_mode->getReportID();
+    buffer.value = INPUT_MODE_TOUCHPAD;
+    buffer.reserved = 0x00;
+    IOBufferMemoryDescriptor* report = IOBufferMemoryDescriptor::inTaskWithOptions(kernel_task, 0, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
+    report->writeBytes(0, &buffer, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
 
+    hid_interface->setReport(report, kIOHIDReportTypeFeature, digitiser.input_mode->getReportID());
+    report->release();
+    
     ready = true;
 }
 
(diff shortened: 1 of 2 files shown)


kprinssu:
This is more or less a revert of https://github.com/VoodooI2C/VoodooI2CHID/commit/c874211f2e15152f2230f98e34d41a740232ddde


I would much prefer to just revert

kprinssu:
@1Revenger1 What do you think of doing the following:

```c++
  void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
      if (MACOS_VERSION > 11) {
         input_mode_element->setValue(INPUT_MODE_TOUCHPAD);
         ready = true;
         return;
      }
      
      // This appears to not work on Catalina or older though (Works in Big Sur/Monterey)
  
      VoodooI2CPrecisionTouchpadFeatureReport buffer;
      buffer.reportID = digitiser.input_mode->getReportID();
      buffer.value = INPUT_MODE_TOUCHPAD;
      buffer.reserved = 0x00;
      IOBufferMemoryDescriptor* report = IOBufferMemoryDescriptor::inTaskWithOptions(kernel_task, 0, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
      report->writeBytes(0, &buffer, sizeof(VoodooI2CPrecisionTouchpadFeatureReport));
  
  
      hid_interface->setReport(report, kIOHIDReportTypeFeature, digitiser.input_mode->getReportID());
      report->release();
  
      ready = true;
  }
}
```

1Revenger1:
I think the change looks good, and can make it later tonight. I still want to figure out why setValue doesn't work, but I don't have a device to test with locally (and I don't want to waste other's time).

kprinssu:
> I think the change looks good, and can make it later tonight. I still want to figure out why setValue doesn't work, but I don't have a device to test with locally (and I don't want to waste other's time).

No worries, I appreciate the effort here. Ideally, we can also use a bit of OOP and IOMatch to selectively load the proper event driver, but we handle that in a follow up PR.

1Revenger1:
@kprinssu this closer to what you were thinking?

kprinssu (approved):

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -12,8 +12,25 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
-    digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);
+    if (version_major > CATALINA_MAJOR_VERSION) {
```
kprinssu, line 15: Just a comment, where is `version_major` coming from? Is it from the `<libkern/version.h>` header?

kprinssu: It is, https://github.com/acidanthera/MacKernelSDK/blob/master/Headers/libkern/version.h#L79

kprinssu:
Thanks @1Revenger1, this solution will work with Catalina and new macOS versions. However, USB trackpad users will need to be on macOS 11+ for VoodooI2C to work.

I think this is an okay compromise.

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
```diff
@@ -23,6 +24,14 @@
 #define INPUT_MODE_MOUSE 0x00
 #define INPUT_MODE_TOUCHPAD 0x03
 
+#define CATALINA_MAJOR_VERSION 19 // Darwin major version for Catalina
+
+typedef struct __attribute__((__packed__)) {
+     UInt8 reportID;
```
ben9923, line 30: A question - `setReport` appears to skip first byte if it equals to the report ID. Why adding it to the buffer, then?

By the HID spec it should be there and I2C-HID just ignores it as it's duplicate?

ben9923 (commented):
Ah, I wish it could be solved with a simple `input_mode->setValue()`. No clue on that, huh?

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
1Revenger1: On some devices, such as @leebinder's, the report ID for input mode is 3. This is same as the trackpad mode, so the mode byte gets skipped.

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -12,8 +12,25 @@
...
+        // Update value from hardware so we can rewrite mode when waking from sleep
+        digitiser.input_mode->getValue(kIOHIDValueOptionsUpdateElementValues);
```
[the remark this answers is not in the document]
1Revenger1: If the value in the input mode element is the same as what we try to set, setValue immediately returns without updating the value in hardware. This updates the element from hardware so that it doesn't immediately return when trying to set a new one.

https://github.com/apple-opensource/IOHIDFamily/blob/master/IOHIDFamily/IOHIDElementPrivate.cpp#L1855-L1861

It will always set the value the first time, which is why we only see this when waking from sleep.

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
kprinssu: We can probably trackdown the issue in a follow up PR. Right now, as it stands, older versions of macOS do not adhere to the HID spec and this is a decent enough work around.

kprinssu:
Thanks @1Revenger1 :)!

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
ben9923: Do you know if calling `input_mode->setValue()` adds the report ID first, so that issue wouldn't happen when we can fully transition to it?

1Revenger1: I'm fairly certain it does. Looking at AlpsHID as well, every report it sends using setReport has the reportID in the first byte as well. This probably means that the USB HID stack assumes that the report ID is in the first byte. Not sure what breaks down in Catalina though.

PR #61, opened 2022-02-25
Clean up various minor memory leaks and small code refactoring to use OSSafeReleaseNULL macro

kprinssu:


ben9923 (approved):
LGTM

BTW, are those leaks caught by the Xcode static analyzer?
Adding it to CI should be quite easy, but we do need to resolve any other issue it raises before adding it.

kprinssu:
These weren't caught, I just took a glance after @mishurov caught an obvious one.

PR #62, opened 2022-02-25
Commented out problematic enterPrecisionTouchpadMode I2C command

kprinssu:
Numerous users are reporting that
`digitiser.input_mode->setValue(INPUT_MODE_TOUCHPAD);` causes trackpads not to function after wake from sleep.

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -89,6 +89,10 @@
 					<integer>13</integer>
 				</dict>
 			</array>
+			<dict>
+				<key>Transport</key>
+				<string>I2C</string>
+			</dict>
 			<key>IOProbeScore</key>
 			<integer>300</integer>
 			<key>IOClass</key>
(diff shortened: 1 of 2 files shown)


ben9923:
With this PR the explicit `setReport` will be used for USB trackpads as well, on every OS version.
We need to check it doesn't break those.

kprinssu:
This is now a partial revert of https://github.com/VoodooI2C/VoodooI2CHID/commit/c874211f2e15152f2230f98e34d41a740232ddde

VoodooI2CHID/Info.plist
```diff
@@ -89,6 +89,10 @@
 					<integer>13</integer>
 				</dict>
 			</array>
+			<dict>
```
ben9923, line 92: Missing `<key>IOPropertyMatch</key>` before it?

kprinssu: Good catch :)! It's been added.

ben9923 (approved):

PR #66, opened 2022-12-09
Fix the kernel panic for USB devices that are connected to Ventura

kprinssu:
This MR is not a guaranteed fix for https://github.com/VoodooI2C/VoodooI2C/issues/507. Custom builds nor support will be provided for this MR. 

Use at your own risk.


peteret:
Hello, could you please turn it into a kext for me?

PR #67, opened 2023-03-28
Working double click for touchscreens

lazd:
This PR enables working double click for touchscreens, such as the ELAN touchscreen on the Microsoft Surface Go 2.

Now you can double click through folders in the Finder with ease, or double click on words to select the word.

In addition to enabling double click, this PR also introduces a method to schedule a lift and adds a constant to define the time to wait between lifts. I increased this value from 14ms to 50ms, which seemed more forgiving when I was working on VoodooI2CGoodix and helps to prevent phantom clicks.

I also cleaned up some of the constants used for click types.

## Todo
* [x] I did notice that selecting a word by double clicking sometimes results in a triple click, so there may be some room for improvement here
* [x] I'm not sure the `millis()` function is needed, I might be able to use `timestamp` instead
* [x] I'm not sure if `cancelTimeout()` is needed in `scheduleLift` -- this was something I did in VoodooI2CGoodix
* [x] Implement a better distance algorithm?

### DIFF
--- a/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
@@ -92,15 +92,21 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
                 isCloseToLastInteraction(x, y) &&
                 (millis() - last_click_time) <= DOUBLE_CLICK_TIME
             ) {
-                // In order to execute a double click, we need to "lift" (hover) where the last tap was, then click again in precisely the same place
-                // Simply change the location we're about the click to where we clicked last and it'll come through as a double click
-                x = last_x;
-                y = last_y;
-                dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, HOVER, x, y);
+                do_double_click = true;
             }
 
             dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, buttons, x, y);
-            
+
+            if (buttons == HOVER) {
+                IOLog("%s::Hover at %d, %d\n", getName(), x, y);
+            }
+            else if (buttons == LEFT_CLICK) {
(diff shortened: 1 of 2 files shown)


VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -167,16 +198,22 @@ void VoodooI2CTouchscreenHIDEventDriver::fingerLift() {
     start_scroll = true;
     uint64_t now_abs;
     clock_get_uptime(&now_abs);
-    
-    dispatchDigitizerEventWithTiltOrientation(now_abs, last_id, kDigitiserTransducerFinger, 0x1, 0x0, last_x, last_y);
-    
+
+    dispatchDigitizerEventWithTiltOrientation(now_abs, last_id, kDigitiserTransducerFinger, 0x1, HOVER, last_x, last_y);
+#ifdef VOODOO_I2C_TOUCHSCREEN_DEBUG
```
kprinssu, line 203: Can we please remove the `VOODOO_I2C_TOUCHSCREEN_DEBUG` preprocessor variable and remove the log as it's device specific?

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -76,22 +70,55 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
...
+            ) {
+                x = last_click_x;
+                y = last_click_y;
+            }
+
+            // Only dispatch a single click event after we've done our hover ticks
+            if ((click_tick <= HOVER_TICKS + 1) || (x != last_x || y != last_y)) {
+                dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, buttons, x, y);
+
+#ifdef VOODOO_I2C_TOUCHSCREEN_DEBUG
```
kprinssu, line 100: Can you also please remove these debug statements too?

kprinssu (approved):

kprinssu:
Thanks for the adding this in @lazd!

PR #68, opened 2023-03-29
Easier to use right click interaction for touchscreens

lazd:
Now you can tap/hold/release and the menu persists or tap/hold/drag/release to select an item.

A timer is now used instead of ticks to determine if a right click should be triggered, and the delay for triggering has been shortened substantially (now it more closely matches Windows).

A deadzone is introduced after right clicking that ensures the menu does not close due to the cursor moving after the right click has been triggered.

After a right click has been triggered, inputs are considered hovers, not clicks, and a special case has been introduced to simulate a click when lifting the finger for a tap/hold/drag/release interaction. This makes using the menu natural for both tap/hold/release/tap to select item in menu or tap/hold/drag/release to select item in menu interaction patterns.

https://user-images.githubusercontent.com/201344/228681320-2ca0ee7d-5b6e-4c94-b3af-6aeb0a90cb11.mov

### DIFF
--- a/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
@@ -62,13 +62,7 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
             ) {
                 right_click = true;
                 dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, RIGHT_CLICK, x, y);
-#ifdef VOODOO_I2C_TOUCHSCREEN_DEBUG
-                IOLog("%s::Right click at %d, %d\n", getName(), x, y);
-#endif
                 dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, HOVER, x, y);
-#ifdef VOODOO_I2C_TOUCHSCREEN_DEBUG
-                IOLog("%s::Hover at %d, %d\n", getName(), x, y);
-#endif
             }
             else {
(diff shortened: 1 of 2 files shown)


VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -39,60 +39,66 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
...
+                touch_start_y = y;
+            } else if (
+                !right_click &&
+                digitiser.contact_count->getValue() == 1 &&
+                isCloseTo(x, y, touch_start_x, touch_start_y) &&
+                (nanoseconds - touch_start_time) >= RIGHT_CLICK_TIME
+            ) {
+                right_click = true;
+                dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, RIGHT_CLICK, x, y);
+#ifdef VOODOO_I2C_TOUCHSCREEN_DEBUG
```
kprinssu, line 65: Can we please remove the debug logs?

kprinssu (approved):

kprinssu:
Thanks again for the hard work @lazd!

lazd:
Thanks, @blankmac! If you get a chance to test this, please let me know if you see any strange behavior and I'd be glad to circle back and fix it. I'll be testing extensively on my machine as well.

lazd:
I'm noticing a couple quirks with this, such as left clicks being triggered before the right click fires, so I'm be working on a fix for that and for scroll behavior sometimes turning into drag if you don't lift both fingers at precisely the same time. If anyone finds other bugs, please let me know!

kprinssu:
@lazd Please create a follow up a PR! Thanks again for the effort.

PR #69, opened 2023-04-01
Prevent phantom clicks when scrolling, clicks when right clicking

lazd:
This PR corrects some issues introduced by #67 and #69 where phantom clicks were happening when scrolling and before right clicking.

The fix is twofold:
1. Move anything that does a left click into a timed function, cancel the clicks if a multitouch interaction starts, don't perform the clicks if a right click has happened (except for the special case of right click + drag)
2. Ignore interactions that happen immediately after a multitouch interaction ends

Because of how this works, `click_tick` is no longer necessary and we're guaranteed to get several hovers before a click or drag happens.

Because drags are delayed, we have to be careful about where we start them, so we use the start location of the touch interaction, which seems to work well.

## Todo
* [x] Figure out what's causing jittering during drags
* [x] Figure out what's cause mouse to stay down after dragging sometimes, resulting in selection
* [x] Figure out why second tap in different location can sometimes turn into drag (scroll, tap, tap)
* [ ] Check stylus behavior due to changes that remove `for` loop from single touch interactions

## Questions about past code that was modified

* [ ] Find out if we need to send `transducer->tip_switch.value()` for finger touches or if `LEFT_CLICK` is fine
* [x] Find out why there was a `for` loop for transducers if there was always one
* [x] Find out why the old `for` loop went to `count + 1`
* [x] Find out if it's ever possible that the fetched transducer is null (could remove checks if not)
* [ ] Double check that returning `false` for `MULTITOUCH_TIMEOUT` is a problem

### DIFF
--- a/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
+++ b/VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
@@ -25,10 +25,10 @@ OSDefineMetaClassAndStructors(VoodooI2CTouchscreenHIDEventDriver, VoodooI2CMulti
 // Override of VoodooI2CMultitouchHIDEventDriver
 
 bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp, VoodooI2CMultitouchEvent event) {
-    absolutetime_to_nanoseconds(timestamp, &last_interaction);
+    absolutetime_to_nanoseconds(timestamp, &last_interaction_time);
     
     // Avoid phantom clicks happening after scrolling
-    if ((last_interaction - last_multitouch_interaction) < MULTITOUCH_TIMEOUT) {
+    if ((last_interaction_time - last_multitouch_interaction_time) < MULTITOUCH_TIMEOUT) {
         return false;
     }
 
@@ -41,7 +41,7 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
         
         if (
             isCloseTo(x, y, last_click_x, last_click_y) &&
-            (last_interaction - last_click_time) <= DOUBLE_CLICK_TIME
+            (last_interaction_time - last_click_time) <= DOUBLE_CLICK_TIME
         ) {
             // Double click start
             // If we're clicking again where we just clicked, precisely position the pointer where it was before
@@ -49,26 +49,26 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
             y = last_click_y;
         }
 
-        if (!finger_down) {
-            finger_down = true;
-            touch_start_time = last_interaction;
+        if (!is_finger_down) {
+            is_finger_down = true;
+            touch_start_time = last_interaction_time;
             touch_start_x = x;
             touch_start_y = y;
         } else if (!is_dragging) {
             bool interaction_is_close_to_start = isCloseTo(x, y, touch_start_x, touch_start_y);
             
             if (
-                !right_click &&
+                !is_right_clicking &&
                  interaction_is_close_to_start &&
-                (last_interaction - touch_start_time) >= RIGHT_CLICK_TIME
+                (last_interaction_time - touch_start_time) >= RIGHT_CLICK_TIME
             ) {
                 // Right click start
                 dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, RIGHT_CLICK, x, y);
                 dispatchDigitizerEventWithTiltOrientation(timestamp, transducer->secondary_id, transducer->type, 0x1, HOVER, x, y);
-                right_click = true;
+                is_right_clicking = true;
             }
 
-            if (right_click) {
+            if (is_right_clicking) {
                 // Continue right click
                 if (interaction_is_close_to_start) {
                     // Adopt the location of the touch start so we don't stray and close the right click menu
@@ -81,9 +81,9 @@ bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp
                 }
             } else {
                 // Check for drag start
-                if (!is_dragging && !is_drag_start_requested && !interaction_is_close_to_start) {
+                if (!is_dragging && !drag_start_requested && !interaction_is_close_to_start) {
                     scheduleDragStart();
-                    is_drag_start_requested = true;
+                    drag_start_requested = true;
                 }
             }
         }
@@ -184,7 +184,7 @@ void VoodooI2CTouchscreenHIDEventDriver::fingerLift() {
     uint64_t now_abs;
     clock_get_uptime(&now_abs);
 
-    if (!is_dragging && (!right_click || moved_during_right_click)) {
+    if (!is_dragging && (!is_right_clicking || moved_during_right_click)) {
         dispatchDigitizerEventWithTiltOrientation(now_abs, last_id, kDigitiserTransducerFinger, 0x1, LEFT_CLICK, last_x, last_y);
     }
     dispatchDigitizerEventWithTiltOrientation(now_abs, last_id, kDigitiserTransducerFinger, 0x1, HOVER, last_x, last_y);
@@ -198,11 +198,11 @@ void VoodooI2CTouchscreenHIDEventDriver::fingerLift() {
 }
 
 void VoodooI2CTouchscreenHIDEventDriver::resetTouch() {
-    finger_down = false;
-    right_click = false;
-    is_dragging = false;
-    is_drag_start_requested = false;
+    is_finger_down = false;
+    is_right_clicking = false;
     moved_during_right_click = false;
+    is_dragging = false;
+    drag_start_requested = false;
 }
 
 IOFramebuffer* VoodooI2CTouchscreenHIDEventDriver::getFramebuffer() {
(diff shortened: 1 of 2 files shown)


VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -190,20 +184,20 @@ void VoodooI2CTouchscreenHIDEventDriver::fingerLift() {
     uint64_t now_abs;
     clock_get_uptime(&now_abs);
 
-    // This seems strange, but we go to hover mode after right clicking, and we need to execute a click if we've moved
-    if (right_click && moved_during_right_click) {
+    if (!is_dragging && (!right_click || moved_during_right_click)) {
         dispatchDigitizerEventWithTiltOrientation(now_abs, last_id, kDigitiserTransducerFinger, 0x1, LEFT_CLICK, last_x, last_y);
```
lazd, line 188: You'll notice that now clicks only happen inside of this timed function. Also, I only ever send `LEFT_CLICK` here, whereas it used to send `transducer->tip_switch.value()`. Do we think that might be an issue?

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -25,67 +25,52 @@ OSDefineMetaClassAndStructors(VoodooI2CTouchscreenHIDEventDriver, VoodooI2CMulti
...
     
-    // If there is a finger touch event, decide if it is single or multitouch.
+    // Avoid phantom clicks happening after scrolling
+    if ((last_interaction - last_multitouch_interaction) < MULTITOUCH_TIMEOUT) {
+        return false;
+    }
+
+    VoodooI2CDigitiserTransducer* transducer = OSDynamicCast(VoodooI2CDigitiserTransducer, event.transducers->getObject(0));
     
-    for (int index = 0; index < digitiser.contact_count->getValue() + 1; index++) {
```
lazd, line 32: Does anyone know why there was a `for` loop here when this function is only ever called for `event.contact_count < 1`?

Also, why did it go to `digitiser.contact_count->getValue() + 1`? It seems like that would always be out of bounds.

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -25,67 +25,52 @@ OSDefineMetaClassAndStructors(VoodooI2CTouchscreenHIDEventDriver, VoodooI2CMulti
...
-    bool got_transducer = false;
+    absolutetime_to_nanoseconds(timestamp, &last_interaction);
     
-    // If there is a finger touch event, decide if it is single or multitouch.
+    // Avoid phantom clicks happening after scrolling
+    if ((last_interaction - last_multitouch_interaction) < MULTITOUCH_TIMEOUT) {
+        return false;
+    }
+
+    VoodooI2CDigitiserTransducer* transducer = OSDynamicCast(VoodooI2CDigitiserTransducer, event.transducers->getObject(0));
```
lazd, line 35: Is it ever possible that the fetched transducer will be `null` here? Previous code assumed that possibility.

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -25,67 +25,52 @@ OSDefineMetaClassAndStructors(VoodooI2CTouchscreenHIDEventDriver, VoodooI2CMulti
...
 
 bool VoodooI2CTouchscreenHIDEventDriver::checkFingerTouch(AbsoluteTime timestamp, VoodooI2CMultitouchEvent event) {
-    bool got_transducer = false;
+    absolutetime_to_nanoseconds(timestamp, &last_interaction);
     
-    // If there is a finger touch event, decide if it is single or multitouch.
+    // Avoid phantom clicks happening after scrolling
+    if ((last_interaction - last_multitouch_interaction) < MULTITOUCH_TIMEOUT) {
+        return false;
+    }
```
lazd, line 33: I return `false` here if the touch happens too quickly after a scroll, which prevents phantom clicks.

However, returning `false` here has the side effect that `multitouch_interface->handleInterruptReport()` will be called. Is that what should happen, or should we return `true` and stop it from being called?

lazd:
@blankmac any insight on the questions I've added as comments on the PR? Thank you!

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
```diff
@@ -249,10 +243,14 @@ void VoodooI2CTouchscreenHIDEventDriver::forwardReport(VoodooI2CMultitouchEvent
...
+            // Immediately cancel any outstanding click
+            timer_source->cancelTimeout();
+            drag_timer_source->cancelTimeout();
+
+            absolutetime_to_nanoseconds(timestamp, &last_multitouch_interaction);
+
             if (event.contact_count == 2 && start_scroll) {
                 scrollPosition(timestamp, event);
-            } else if (event.contact_count == 2 && !start_scroll) {
-                scheduleLift();
```
lazd, line 255: I'm wondering if this might cause a problem. I removed it since lifts are now also clicks, but maybe it's required?

blankmac:
I think you're finding out that making what amounts to a mouse or touchpad input behave like you'd expect a touchscreen to behave is a complicated thing!  Unfortunately, I no longer have any hackintosh hardware with a touchscreen.  With Apple moving towards their own CPUs I bit the bullet and purchased an M1 based laptop when they came out.  I have a friend that still has an HP Elite X2 tho, which is what I originally used to write the touchscreen routines, and will get it from him in a few days and try your revisions.  It's been so long since I've dug into this code that there are most certainly redundant portions and errors.  It definitely needs an update!

Regarding the for loop and the "+1", IIRC that has to do with the transducer ordering.  The stylus always comes first but I believe the catch was that it wouldn't appear in the transducers as a contact event until you had actually used it once.  You might try using a stylus first and then seeing if your single touch / multi touch input behaves as expected.

It's unfortunate that so much complexity has to be added for such a simple thing as double clicking.  Through 10.11, double clicking worked just fine and as expected.  There's an area property in the HID device that I believe is the root of the problem but I was never successful at widening the area to allow for slightly different locations when executing a double click.  

Anyway, I should be able to go through your revisions in a couple days.  But don't let that stop you if you get it figured out prior to that.  Thanks for working on it!

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
lazd: It was! This was how the driver knew that a scroll ended. Without this, it didn't position the cursor between the two fingers when a scroll started. I introduced a new timer to handle this.

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
lazd: It seems to be due to the +1 in the for loop.

lazd:
@blankmac thanks for the insight! I've purchased a Microsoft Surface Pen so I can test stylus interactions. I'll check and see if that `for` loop +1 index is still required.

lazd:
@kprinssu actually yeah, I can't get my Surface Go 2's touchscreen to react to a pen in macOS, so there's no way for me to test if pen support is still behaving as expected. I believe that's the only question that remains, the other two items under `Questions about past code that was modified` aren't as pressing.

If touch works, this could be merged.

kprinssu:
Thanks @lazd, I'll take a look sometime this week and will add my comments to this PR.

lazd:
@kprinssu I'd say go ahead and get this in for the 2.9 release, this fixes some behavior from my previous PRs. Let's see what users say, and if we broke pens I can fix it. Thank you!

kprinssu:
Thanks @lazd

kprinssu (approved):

zhen-zen:
Was too busy to debug my tablet recently. I tried the latest master and the touchscreen started to behave like a big touchpad. Single touch won't move the cursor to finger as usual, and swiping translates to relative movement of the cursor. The first commit in this PR started to cause the issue but I didn't dig into the code further.

lazd:
> Was too busy to debug my tablet recently. I tried the latest master and the touchscreen started to behave like a big touchpad. Single touch won't move the cursor to finger as usual, and swiping translates to relative movement of the cursor. The first commit in this PR started to cause the issue but I didn't dig into the code further.

Interesting, I'm running this on my Surface Go 2 and the touchscreen works great. It sounds like the driver isn't even being loaded for your touchscreen since it's being treated like a trackpad, perhaps there is a configuration error?

zhen-zen:
I don't think there's any change in configuration since Info.plist is untouched. Build of previous commit just works and the HEAD breaks one finger tracking.

image

zhen-zen:
Was able to dig into the code flow a bit, thanks to the cloudy weather today that blocks the ring of fire :(
I am suspecting `checkFingerTouch` as the report will be hand to `multitouch_interface` so it will then behave as a touchpad. Since IOLog won't work under interrupt context, I used setProperty to indicate the return of this method. And after the PR, it always return false. Not sure if I still have some time to continue debugging over the weekend.

VoodooI2CHID/VoodooI2CTouchscreenHIDEventDriver.cpp
zhen-zen: Here seems to be the culprit. Finger is not always the first transducer. It seems that pen is the first one on my WCOM48CE.

PR #71, opened 2023-05-10
Use setValueData for USB Trackpads

1Revenger1:
`setValueData` has no checks for duplicate data, so I'm hoping will work after sleep and in older versions of macOS.
I've had someone test this to work in Catalina on a USB trackpad, but this probably requires more testing before being merged in. I'd rather let this sit to make sure it doesn't have issues before merging it in.

Tested Machines:
- Acer laptop with USB type cover (macOS 10.15)
  - Only issue is that VoodooI2CHID got unloaded if the type cover was disconnected for a few minutes :(
- Zenbook Duo UX481FA (I2C, macOS 12)

### DIFF
--- a/VoodooI2CHID/Info.plist
+++ b/VoodooI2CHID/Info.plist
@@ -89,11 +89,6 @@
 					<integer>13</integer>
 				</dict>
 			</array>
-			<key>IOPropertyMatch</key>
-			<dict>
-				<key>Transport</key>
-				<string>I2C</string>
-			</dict>
 			<key>IOProbeScore</key>
 			<integer>300</integer>
 			<key>IOClass</key>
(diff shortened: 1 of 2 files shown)


kprinssu:
This will be a tough to test without community help and pre-built binaries.

1Revenger1:
[VoodooI2CHID.kext.zip](https://github.com/VoodooI2C/VoodooI2CHID/files/11466573/VoodooI2CHID.kext.zip)
Oops, I suppose I could at least provide a binary.

kprinssu:
@1Revenger1 What do you think of me merging this in and getting into the 2.9 release?

We can always revert this change back if we run into any issues.

zhen-zen:
I wish I could have read your PR earlier. USB touchpad only works in mouse mode and I have to limit the driver to I2C transport so it won't stuck. Thought there was touchpad FW issue at the beginning since input mode was already set.

It should fix at least 44e:1218 typecover i.e. https://github.com/blankmac/AlpsHID/issues/20

kprinssu:
@1Revenger1 would you be able to rebase or fix the merge conflicts for this PR?

I would like to merge this in as it is.

PR #72, opened 2023-08-07
Always parse pen usage

zhen-zen:
Fix broken pen input in parseElements filter.


kprinssu:
Great work as usual @zhen-zen

kprinssu (approved):

PR #75, opened 2024-06-11
Trackpoint

1Revenger1:
Looks for pointer device reports and parses them when the device sends those reports. On hold for Sequoia shenanigans


Lorys89:
Send kext for test

1Revenger1:
It causes panics for some reason registering interrupts in VoodooI2CELAn. I pushed partially so I could wipe my local repo haha.

kprinssu:
@Lorys89 Please do not spam Github PRs. I have temp banned you for a bit as you have done this before.

kprinssu:
@1Revenger1 More than likely there are breaking changes and we need to update VoodooInput upstream.

1Revenger1:
Yep, we figured it out in Discord. Looks like they removed the IOKit personality for the product id 0x272. Setting it to 0x265 to match the MT2 seems to work just fine. Not entirely sure if that is the best solution

kprinssu:
We'll need to create a PR upstream in VoodooInput.

PR #78, opened 2026-04-17
Fix Precision Touchpad wake race condition with delayed retry scheduling

bliss95:
Fix Precision Touchpad wake race condition after sleep.

Some slower systems fail to initialize digitiser.input_mode immediately after wake, causing a kernel panic in enterPrecisionTouchpadMode().

This patch introduces:

- delayed retry scheduling using thread_call
- retry limit protection
- proper cleanup in stop()
- readiness guard before handling reports

Tested on:
ASUS UX430UNR
Intel I7-8550U as MacBookPro16,2
macOS Tahoe 26.4.1
VoodooI2C 2.9.1

Fixes panic occurring at:
VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode()

Before:
Wake Event
   │
  ├─ setPowerState()
   │
  ├─ enterPrecisionTouchpadMode()
   │        │
   │       ├─ input_mode NOT READY ❌
   │      └─ NULL dereference → PANIC
   │
  └─ device init (too late)

After:
Wake Event
   │
  ├─ setPowerState()
   │
  ├─ ready = false
   │
  ├─ interrupt arrives
   │
  ├─ enterPrecisionTouchpadMode()
   │        │
   │       ├─ input_mode not available → schedule retry
   │        │
   │       ├─ wait 50 ms
   │        │
   │       ├─ retry initialization
   │        │
   │       └─ input_mode becomes available → success
   │
  └─ Trackpad becomes operational


kprinssu:
I appreciate the PR and attempting to fix this race condition.


I don't think this is a good idea adding as we now have the added complexity of creating and managing a lifecycle of a thread. I suggest we to stick to IOKits established paradigms.

I would suggest us to do the following:
- add checks to avoid dereferencing the variable 'value' when it is NULL
- remove the explicit thread creation
- use a simple backoff algorithm (keep the retries, use IOSleep to ensure the I2C bus is ready)

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -12,10 +12,32 @@
...
+        return;
+    }
+    
     UInt8 inputMode[] = { INPUT_MODE_TOUCHPAD };
     
     // Use setDataValue as it does not check for duplicate writes
     OSData *value = OSData::withBytes(inputMode, sizeof(inputMode));
+    
+    if (!value) {
+        IOLog("VoodooI2CHID: failed to allocate OSData\n");
```
kprinssu: Suggestion: can we add a better log message? Such as "VoodooI2CHID: failed to allocate memory for input mode setDataValue"

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -12,10 +12,32 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
+    const int maxRetries = 10;
+    int retry = 0;
+
+    // Wait until input_mode becomes available
+    while (!digitiser.input_mode && retry < maxRetries) {
+        IOLog("VoodooI2CHID: input_mode not ready, retry %d\n", retry);
```
kprinssu: Suggestion: I think these logs are redundant and will be quite spammy. Can we consider removing this?

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.cpp
```diff
@@ -12,10 +12,32 @@
 OSDefineMetaClassAndStructors(VoodooI2CPrecisionTouchpadHIDEventDriver, VoodooI2CMultitouchHIDEventDriver);
 
 void VoodooI2CPrecisionTouchpadHIDEventDriver::enterPrecisionTouchpadMode() {
+    const int maxRetries = 10;
```
kprinssu: Suggestion: Can we move this constant into the header and make it static?

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
```diff
@@ -60,7 +63,13 @@ class EXPORT VoodooI2CPrecisionTouchpadHIDEventDriver : public VoodooI2CMultitou
     IOReturn parseElements(UInt32) override;
  private:
     bool ready = false;
-
+    
+    //
+    int retry_count = 0;
+    bool retry_scheduled = false;
+    uint64_t last_retry_time = 0;
+    thread_call_t retry_call = nullptr;
```
kprinssu: I don't think we need this variable anymore?

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
```diff
@@ -60,7 +63,13 @@ class EXPORT VoodooI2CPrecisionTouchpadHIDEventDriver : public VoodooI2CMultitou
     IOReturn parseElements(UInt32) override;
  private:
     bool ready = false;
-
+    
+    //
```
kprinssu: Nitpick: can we remove this empty comment?

VoodooI2CHID/VoodooI2CPrecisionTouchpadHIDEventDriver.hpp
```diff
@@ -60,7 +63,13 @@ class EXPORT VoodooI2CPrecisionTouchpadHIDEventDriver : public VoodooI2CMultitou
...
+    int retry_count = 0;
+    bool retry_scheduled = false;
+    uint64_t last_retry_time = 0;
```
kprinssu: I don't think variable is used.

kprinssu:
@bliss9 I was chatting w/ @averycblack and we were wondering if you have a USB trackpad. As we think the retry fix shouldn't be needed here.

bliss95:
Thanks for the suggestion.

I tested this on an ASUS UX430UNR using the built-in I2C touchpad.

One important observation:
Sleep/Wake works correctly with VoodooI2C v2.8 on this machine.
However, with the current version, the internal I2C trackpad fails after sleep/wake,
and this results in a kernel panic.

The attached panic log confirms the failure during initialization.

This looks like a regression introduced in newer versions.

I do not currently have a USB trackpad available to compare behavior,
but the issue consistently reproduces on the internal I2C device.

The retry logic was added because input_mode was occasionally not ready
when enterPrecisionTouchpadMode() was called after wake.

kprinssu:
Thanks again for the fixes. I am hoping another person to can help test the changes.

I will be going on vacation shortly and will not have a hackintosh in hand to test this with. The code looks good to me.

